PR #2, opened 2024-01-26
feat: OpenAPI proposal for V1 OFREP

Kavindu-Dodan:
## OFREP evaluation APIs

A proposal for the initial version of the OpenAPI specification for the OFREP service. 

Please feel free to review, suggest, and propose changes so we can collectively define the service. 

### Why OpenAPI?

According to the survey results, JSON over HTTP was the most widely used technology by vendors.

### What's in this PR? 

image

This PR mainly focuses on the flag evaluation. It contains,

- Endpoint for individual flag evaluation - `/ofrep/v1/evaluate/{key}:`

This endpoint evaluates an individual feature flag from the flag management system

- Endpoint for bulk evaluation - `/ofrep/v1/evaluate`

This endpoint mainly focuses on the [static context paradigm](https://openfeature.dev/specification/glossary/#static-context-paradigm) and allows multiple flags (bulk) evaluations based on context, authorization token, or flag keys in the request payload. 

This PR does NOT focus on telemetry or flag evaluation metrics. This will be a follow-up discussion.


service/openapi.yaml
```diff
@@ -0,0 +1,285 @@
...
+                example: '2024-02-07T12:00:00Z'
+        '500':
+          description: Internal server error
+  /ofrep/v1/flag/changes:
+    post:
+      description: OFREP polling endpoint to check if the flags have been modified
+      requestBody:
+        content:
+          application/json:
+            schema:
```
nicklasl, line 104: Not 100% sure how this endpoint will work. Let me know if I've understood correctly:

The provider can use this endpoint to understand if the flag "rules" in the flag provider system (backend) have changed since the providers last evaluation and decide to make an evaluate call even if the evaluation context has not changed?

thomaspoignant: Yes sorry if the context is not clear, but this is exactly that.

The goal of this endpoint is to do polling on it if the `ETag` is different than the one in the cache of the provider we will call the `evaluate` endpoint for the changed flags.

It avoids having to go through re-evaluation too often.
**Why are we doing polling?** Because this is the most voted one in the OFREP survey, but in the future we should also probably support Server Send Event or Websocket

thomaspoignant: If you want to have a look I've started a very draft PR of an OFREP web provider https://github.com/open-feature/js-sdk-contrib/pull/776.

It may help you to see how we use this endpoint.

jonathannorris: Why does this need to be a different endpoint? Couldn't you set the `ETag` header on any of the other endpoints and have them return `304` responses? 

I guess it depends on how the provider is implementing `ETag` headers for these endpoints, but I assume they would be a hash on the FF configuration for a user, and not the hash of the underlying project configuration. Because in most platforms flag evaluation can be dynamic based on things like schedules that wouldn't be captured by the project configuration `ETag`.

service/openapi.yaml
```diff
@@ -0,0 +1,245 @@
...
+  securitySchemes:
+    BearerAuth:
+      description: (optional) Bearer Authorization to your flag management system.
+      type: http
+      scheme: bearer
+    ApiKeyAuth:
+      description: (optional) API Key to your flag management system.
+      type: apiKey
+      in: header
+      name: X-API-Key
```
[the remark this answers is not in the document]
jonathannorris: yea this seems reasonable to me.

service/openapi.yaml
```diff
@@ -0,0 +1,285 @@
...
+          metadata:
+            type: object
+            description: Arbitrary metadata supporting flag evaluation
+          errorCode:
+            type: string
+            description: OpenFeature compatible error code. See https://openfeature.dev/specification/types#error-code
+          errorDetails:
+            type: string
+            description: Optional error details description for logging or other needs
+          ETag:
```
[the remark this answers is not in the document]
jonathannorris: Shouldn't this be in the response headers, not in the payload body?

service/openapi.yaml
```diff
@@ -0,0 +1,87 @@
...
+              schema:
+                $ref: "#/components/schemas/failure"
+components:
+  schemas:
+    evaluationRequest:
+      description: Flag evaluation request
+      properties:
+        flagKey:
+          $ref: "#/components/schemas/flagKey"
+        flagType:
```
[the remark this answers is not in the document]
jonathannorris: Interesting discussion here. I think it really comes down to where do we put the responsibility of the type miss-matches getting resolved. Seeing that all the OpenFeature SDKs have typed flag methods `getBooleanValue()`, and that this API is basically a remote-evaluation interface for a Provider, and all the provider methods are typed `getStringEvaluation()` and expected to handle the type miss-matches, I think we should have these requests typed in some form. 

And seeing that we want to have a bulk evaluation endpoint, it doesn't make sense to have the API routes be individually typed and have to handle the bulk endpoint differently. Therefore the `evaluationRequest` should have a required type on it.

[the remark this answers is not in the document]
lukas-reining: > The concern I have is the knowledge of the type at flag management system. I can speak for flagd, where flag schema doesn't enforce flag type.

@Kavindu-Dodan 
But that's exactly why I do not see the need for a type in the request but rather in the response.
The flag management will have to have any knowledge of which type it has evaluated to. And maybe there is a case where it does not know for some reason but then the type in the response could be optional.
The provider can then do look at the type and try to parse the object value into the one specified in the response or declare a TYPE_MISMATCH if the evaluation method was called for another type.


> Seeing that all the OpenFeature SDKs have typed flag methods getBooleanValue(), and that this API is basically a remote-evaluation interface for a Provider, and all the provider methods are typed getStringEvaluation(), and the providers are expected to handle the type miss-matches, I think we should have these requests typed in some form.

I get the point.
But having the type in the request, or even type specific endpoints somewhat implies that flags could have more than one type. To me it would be way clearer to just evaluate a flag by a key and contextual information and then check the type based on the (optional) type in the response and the result of parsing as a consumer, which in our case is the provider implementation.


This is a long discussion and I am not sure how to get to a point here actually😅

service/openapi.yaml
thomaspoignant: The main goal of having a different endpoint was to avoid having to re-evaluate all the flags, but your point about evaluation being dynamic is super interesting.

I am not opposed to regularly re-evaluating and sending the ETag back _(just to ease the comparison)_,  but this can put pressure on the different flag management systems.

About the 304, how do you imagine doing it on the server side? Because it does not have the context of the provider, do you expect the flag management systems to keep some state?

service/openapi.yaml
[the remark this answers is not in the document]
thomaspoignant: We have put it in the body because we have 1 ETag per flag to avoid re-fetching/re-evaluating the full list if only 1 flag has changed.

I am open to alternative ideas here.

service/openapi.yaml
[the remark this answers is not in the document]
thomaspoignant: I tend to agree with @lukas-reining, especially since the bulk evaluation can be "evaluate everything", so no type of definition on the client.

And the flag management systems knows what are the types of the flags, so if the flag comes from the response it makes it explicit and the provider can just work with the type on his own.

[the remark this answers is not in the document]
beeme1mr: I'm in favor of not including the expected type in the request but it should be included in the response body. That would work with both bulk evaluations and individual evaluations.

service/openapi.yaml
```diff
@@ -0,0 +1,87 @@
...
+          description: Context information for flag evaluation
+      required:
+        - flagKey
+        - flagType
+    success:
+      description: Flag evaluation success response.
+      properties:
+        flagKey:
+          $ref: "#/components/schemas/flagKey"
+        value:
```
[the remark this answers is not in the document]
beeme1mr: I believe the type should be included in the response.

service/openapi.yaml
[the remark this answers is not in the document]
beeme1mr: The ETag should either be used in the header or we should come up with a different term. I would try to leverage ETags (as a header) if possible because it has broad compatibility and is built for caching use cases.

I don't think it's necessary to support change detection on each flag during a bulk evaluation. That would simplify the implementation and allow for the ETag to be on the response header.

service/openapi.yaml
beeme1mr: I would like to avoid adding a different endpoint if possible. As @jonathannorris mentioned, we should be able to reuse the same API and return a `304` if the ruleset hasn't changed or updated context hasn't affected the evaluated response. This would require context to be sent in every request but the overhead should be minimal.

thomaspoignant: Do we really want to re-evaluate everything all the time?

service/openapi.yaml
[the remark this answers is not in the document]
kinyoklion: I am in favor of a type in the request. Primarily because I think the provider for a specific vendor should operate effectively the same with remote or local evaluation. Requests are made to the provider for specific types currently, so I think they also should be remotely.

In addition it also means that the provider has visibility. In the LaunchDarkly case each flag that is evaluated with a type that cannot be converted is also going to appear in analytics. It cannot appear if the provider isn't making that determination.

I am not sure what we are going to do about the evaluating all flags case. It seems to be a problem with both local and remote evaluation. I think the reasonable approach is to use an intermediate representation, most similar to a `structure`  for OF purposes or just an `any` type in languages that support it. All flags can be the type, and then the caller has to handle conversion (just keeping in mind that they are not always a 1-to-1 situation, especially numbers).

service/openapi.yaml
jonathannorris: I kind of question the value of the `ETag` header here at all; for most providers with some dynamic evaluation, we'd likely be doing a complete evaluation every time anyways. Therefore, the only thing we'd be saving is the bandwidth of returning the flag results in a `200` response vs a `304` response. 

Maybe letting providers set `max-age` headers in their response and using standard caching headers from there may let providers control how often they want to re-evaluate flags. But from DevCycle's perspective, we'd just re-evaluate every time; we have a dynamic evaluation at the edge where we can merge client-side contexts with our edgeDB user database, in addition to schedules and other rollout mechanics that should be evaluated every time. Also, a user evaluation is usually less than the overhead of accepting the HTTP request.

service/openapi.yaml
```diff
@@ -0,0 +1,285 @@
...
+      type: array
+      description: (optional) List of flags to be evaluated, if not set all the flag available will be evaluated against the context.
+      items:
+        $ref: '#/components/schemas/key'
+      example:
+        - my-flag
+        - my-other-flag
+    context:
+      type: object
+      description: Context information for flag evaluation
```
toddbaert, line 269: Is this optional? My OpenAPI is a bit rusty... I would encourage that we make it optional, I think, consistent with the OpenFeature API.

service/openapi.yaml
toddbaert: > Also, a user evaluation is usually less than the overhead of accepting the HTTP request.

I suspect this is generally the case. I think opening a TCP connection, securing it, and  deserializing a JSON payload is probably more work than evaluating tens of flags in most systems.

> I kind of question the value of the ETag header here at all; for most providers with some dynamic evaluation, we'd likely be doing a complete evaluation every time anyways. Therefore, the only thing we'd be saving is the bandwidth of returning the flag results in a 200 response vs a 304 response.

If you have context-insensitive flags (a simple rollout flag that's on/off for everyone, for example) and `ETag`/`If-None-Match` could be valuable; in that case if the provider sends `If-None-Match: abc123` and that matches the currrent `ETag`, we know we can actually skip evaluation.

> Maybe letting providers set max-age headers in their response and using standard caching headers from there may let providers control how often they want to re-evaluate flags.

I like this idea.

I think overall we can get the optimal caching behavior we want without the dedicated `/change` endpoint. For an implementation to determine if it should send a 304, it would check the `If-None-Match` header and also determine if any of the flags implicated in the request are context sensitive. Additionally, if no context is included in the request, it might be safe for an implementation to assume it can send a 304.

service/openapi.yaml
```diff
@@ -0,0 +1,285 @@
...
+          headers:
+            Retry-Later:
+              description: Indicates when to retry the request
+              schema:
+                type: string
+                format: date-time
+                example: '2024-02-07T12:00:00Z'
+        '500':
+          description: Internal server error
+  /ofrep/v1/flag/changes:
```
toddbaert, line 98: I think we should avoid spec'ing this endpoint, _at least_ for now. See my comment [here](https://github.com/open-feature/protocol/pull/2#discussion_r1496311806).

service/openapi.yaml
```diff
@@ -0,0 +1,285 @@
...
+  description: OFREP define the protocol for remote flag evaluations
+  contact:
+    url: https://github.com/open-feature/protocol
+  license:
+    name: Apache 2.0
+security:
+  - ApiKeyAuth: []
+  - BearerAuth: []
+paths:
+  /ofrep/v1/evaluate/{key}:
```
toddbaert, line 14:
```suggestion
  /ofrep/v1/evaluate/flags/{key}:
```

I'm not positive, but this might afford clearer semantics, and be more future-proof. For example, if we add some kind of concept for a collection or subset of flags (let's call it a scope) you might be able to evaluate just that subset with `/ofrep/v1/evaluate/scopes/{scope-id}`.

An alternative to consider would be to support some kind of query-param based filtering on the bulk endpoint:

`/ofrep/v1/evaluate?scope=myScope`

Neither of these need to be implemented now, just food for thought.

service/openapi.yaml
Kavindu-Dodan: Here we only define reusable schemas. In OpenAPI unless stated all parameters are optional. We have to use `required` keys if we enforce mandatory fields. See this for example [1]

[1] - https://github.com/open-feature/protocol/blob/0e4d0ddf4cc3b953695d8bd63670f95420b3d139/service/openapi.yaml#L228-L230

service/openapi.yaml
thomaspoignant: I like the addition of `/flags/`, I will add it.

For the query params or any scope-related thing, I agree that this is something that can come later.

service/openapi.yaml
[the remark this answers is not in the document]
beeme1mr: Update: I don't think the type is necessary in the request or response. Developers must specify the expected type when evaluating a feature flag. The provider can then perform a runtime type check. I've opened up a separate PR containing the proposed changes.

service/openapi.yaml
thomaspoignant: I am fine to go for re-evaluation if we think this is the way to go.

So I try to sum up the discussion here:
- We can get rid of the `/change` endpoint.
- It is fine to ask for re-evaluation each time.
- We will have an `ETag` as a header that is computed for the full request _(not per flag)_.

Is this what you had in mind @toddbaert @jonathannorris ?

thomaspoignant:
**I've added some changes:**
- remove the `/changes` endpoint
- add a `ETag` header in the bulk evaluation success response
- add the type in the response of both evaluation APIs, based on the @lukas-reining proposal.
- some cleanup

service/openapi.yaml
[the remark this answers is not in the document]
thomaspoignant: As I said in the beginning the problem is more with double vs int because JSON does not make the difference and this is where it becomes tricky.

For example, in GO all integers are marshall in float which makes it complex to handle properly the TYPE_MISMATCH between those 2 types, we are kind of lost in the JSON translation.
If we request to be explicit on the type from the flag management system, we will reduce the choice that the provider should take.
I think that having the type in the response is a good approach here and it will simplify a lot how to build providers.

The example I have in mind is, that you can receive a int that looks like:
```json
...
"value":1.0
...
```

How do you know if it is an `int` or a `float`? We are missing context in the provider to know if we should return a `TYPE_MISMATCH` error or if we should cast the value into an `int`.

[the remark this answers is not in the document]
kinyoklion: To add on to this. Even if, at one point in time, you get a value that looks like `"value":1`, then treating that as specifically an integer can still cause you trouble.

The consumer of the code handles it as an integer only, and then someone changes the value to `0.5` and now the type becomes a double and things break.

"I want this thing, and I want it as an int" is a safer interface.

It doesn't happen as often, but this is also true for other types. If a flag was a string, and it gets changes to an int, then the interface of "get me this thing as a string" can still provide a correct default value and indicate a `TYPE_MISMATCH`, and it can be done at the provider level, which means vendor analytics are still correct.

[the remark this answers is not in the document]
thomaspoignant: The tricky part of asking in the API request will be that we are not always requesting a flag, in the bulk evaluation we can ask for everything available.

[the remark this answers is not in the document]
kinyoklion: If you don't want to transmit it in the request, and you just want to get whatever type it is back, then likely you need another bit of communication. Something that lets the provider know the outcome. The type wasn't what was expected and we used the default value. (Though this is a harder sell because now the vendor SDK needs a way to ingest that information as well.)

As it currently is I think the default value information is missing as well? I have concerns that the fidelity of information getting back to the provider is going to hinder things.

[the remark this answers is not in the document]
thomaspoignant: The idea is to send back the type configured from the flag in the response, so it is explicit and in the provider you can compare the method called and the type configured in flag management system.

Let's say that you configure a flag as double in LD, the bulk API response will be something like this:
```json
[
  {
    "type": "double",
    "value": 1,
    "key": "my-flag",
    "reason": "TARGETING_MATCH",
    "variant": "A",
  },
  {
    "type": "string",
    "value": "my-flag-value",
    "key": "my-flag",
    "reason": "TARGETING_MATCH",
    "variant": "B"
  }
]
```

With the `type` you know what is expected, allowing you to cast the `1` into a double with no doubt because you know the type of the flag.

It means that if you call `client.getIntegerValue("my-flag", 5.1);` you should expect an errorCode `TYPE_MISMATCH` even if in the JSON format it looks like an integer.

service/openapi.yaml
toddbaert: I think that's the right direction for an initial version.

service/openapi.yaml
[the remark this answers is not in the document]
kinyoklion: If I was using this I would want to know that using the remote evaluation is equivalent to in-process evaluation in terms of results and ideally analytics. I don't think that will be true with this approach, it reverses the driver of typing.

An individual evaluation call through remote evaluation may produce different results, have different constraints, and different produce different analytics.

For example the LD case would encounter at least these:
1.) Only double support for evaluation methods.
2.) The provider wouldn't have a way to produce wrong type analytics.
3.) The provider wouldn't have a way to provide default values in analytics.

So if you use remote eval you are basically getting a different product, not just out of process eval.

service/openapi.yaml
```diff
@@ -0,0 +1,87 @@
...
+        value:
+          type: object
+          description: Flag evaluation result
+        reason:
+          type: string
+          description: Reason matching OpenFeature reason
+        variant:
+          type: string
+          description: Variant of the evaluated flag value
+        metadata:
```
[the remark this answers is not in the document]
dyladan: It seems to me like there is a ton of overlap between bulkEvaluationSuccess and evaluationSuccess. I wonder if the bulkEvaluationSuccess could be simplified to:

```yaml
      description: Flag evaluation success response.
       allOf:
         - properties:
             flags:
               type: array
               items:
                 - anyOf:
                   $ref: "#/components/schemas/evaluationSuccess"
                   $ref: "#/components/schemas/evaluationFailure"
```

service/openapi.yaml
```diff
@@ -0,0 +1,294 @@
...
+                description: Optional error details description for logging or other needs
+            required:
+              - key
+              - value
+          - oneOf:
+              - $ref: "#/components/schemas/booleanFlag"
+              - $ref: "#/components/schemas/stringFlag"
+              - $ref: "#/components/schemas/integerFlag"
+              - $ref: "#/components/schemas/floatFlag"
+              - $ref: "#/components/schemas/objectFlag"
```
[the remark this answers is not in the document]
Kavindu-Dodan: @beeme1mr  converted bulk evaluation response to an object. @dyladan  Combined schemes such a way that we now have `oneOf` (instead of `anyOf`) - This simplifies schema to have an individual flag evaluation as success or failure

I will resolve this conversation now

service/openapi.yaml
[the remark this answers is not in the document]
Kavindu-Dodan: Will resolve as we followed up with the suggestions (and see [here](https://github.com/open-feature/protocol/pull/2#discussion_r1503180581) )

Kavindu-Dodan:
**Updates (26.02.2024)**

- Bulk evaluation success response is now an object enclosing the flags array
  - An individual flag response in bulk evaluation can be either a success or a failure
  - Reuses schemas from single flag evaluation 

<details>

<summary>Preview bulk evaluation response example</summary>


```json
   {
    "flags": [
      {
        "value": true,
        "key": "my-flag",
        "type": "boolean",
        "reason": "string",
        "variant": "string",
        "metadata": {}
      },
      {
        "key": "my-flag",
        "reason": "string",
        "variant": "string",
        "metadata": {},
        "errorCode": "FLAG_NOT_FOUND",
        "errorDetails": "string"
      }
  ]
}
```
</details>

- Removed 404 flag not found as we already cover this with 400 response encapsulating all error codes 
- Improved flag type


**Open discussions** 

- Usage of type parameter in request & response 
- Usage of ETag in response headers

service/openapi.yaml
```diff
@@ -0,0 +1,246 @@
...
+        context:
+          $ref: '#/components/schemas/context'
+    bulkEvaluationSuccess:
+      description: Success response for the bulk evaluation request
+      type: object
+      properties:
+        flags:
+          type: array
+          items:
+            oneOf:
```
dyladan, line 126: I believe if you use `oneOf` it means all items in the list must have the same type. Isn't it possible to have a bulk where one flag is successful and another flag is a failure? I believe you need to use `anyOf` for that.

Kavindu-Dodan: @dyladan Yes, the bulk response should contain both success and failures. And I interpreted this rule to be an _array containing items that match oneOf the schemas from success or failuire_. I couldn't find a lot of resources on correct interpretation when oneOf is used within an array. But I came across this discussion [1] which suggests this is correct. 

[1] - https://community.smartbear.com/discussions/swaggerhub/the-use-of-oneof-or-anyof-in-arrays/257567

thomaspoignant:
After the community meeting, I pushed the last decision made.

For context here is what we have discussed:
We end up by simplifying quite a bit the spec while keeping in mind that we can always add complexity later.
- In order to minimize serialization and ETag management, we will add the `if none match` header support in the requests.
- We've decided to remove any notion of types in spec, we will rely only on JSON type and we will have to cast the data when needed (mostly between int and float).
- We've decided to remove the flags field in the bulk evaluation, if for any reason a flag management system wants to have a subset of the flags it will be managed in the context.

**Next steps:**
- ✅ We will update the openAPI spec and try to merge the PR https://github.com/open-feature/protocol/pull/2
- We will write some ADRs to track the decisions taken.

**After the PR is merged we have a few topics to discuss such as:**
- provider metadata
- telemetry

service/openapi.yaml
```diff
@@ -0,0 +1,239 @@
+openapi: 3.1.0
+info:
+  version: 0.1.0
+  title: OpenFeature Remote Evaluation Protocol (OFREP)
+  description: OFREP define the protocol for remote flag evaluations
+  contact:
+    url: https://github.com/open-feature/protocol
+  license:
+    identifier: Apache 2.0
```
beeme1mr, line 9:
```suggestion
    identifier: Apache-2.0
```

This must match the SPDX identifier defined [here](https://spdx.org/licenses/).

beeme1mr (approved):

lukas-reining:
# ADRs

As discussed in the OFREP meeting and with @Kavindu-Dodan and @thomaspoignant I added ADRs for the most decisions that we felt like are important to track and have a record of.

We came up wit the following points to track in an adr:

- HTTP / JSON
- Polling over other systems
- Re-evaluation for each polling
- No explicit type handling

The structure of ADRs is bases on Michael Nygards template: [https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions)
For managing ADRs we use [https://github.com/npryce/adr-tools](https://github.com/npryce/adr-tools).
The ADRs can still be created and managed by hand, the tool just helps creating new ones and linking them so it is optional to use, while the structure of each file and naming should followed.

lukas-reining (approved):

thomaspoignant (approved):

service/adrs/0002-two-evaluation-endpoints.md
```diff
@@ -0,0 +1,28 @@
...
+## Decision
+
+OFREP contains two endpoints, one for individual flag evaluations and the other for bulk evaluations.
+Individual flag evaluation is preferred for the dynamic context paradigm whereas the bulk evaluation is preferred for static context paradigm.
+
+## Consequences
+
+- OFREP supports single and bulk evaluations
+- Each endpoint might evolve to better serve the specific OpenFeature paradigm
+- It might become challenging to prevent the endpoint from diverging
```
toddbaert, line 28:
```suggestion
- It might become challenging to prevent the endpoints from diverging
```

service/adrs/0003-no-flag-type-in-request-or-response.md
```diff
@@ -0,0 +1,33 @@
...
+Details can be seen in the [OF specification](https://openfeature.dev/specification/types).
+
+When evaluating a flag via OFREP, the provider implementation will have to parse the flag value from the JSON response.
+There are different approaches to this. Some feature flagging services define an expected flag type in the request.
+Some include the flag type or even a schema in the response and some do not specify any type either in request or response.
+
+## Decision
+
+In the community meeting on 2024-02-29, it was concluded **not** to include type details in any payload.
+The reason for this is to keep the API simple till there is a need for improvements.
```
toddbaert, line 22:
```suggestion
The reason for this is to keep the API simple until there is a need for improvements.
```

jonathannorris (approved):

toddbaert:
Nice work @Kavindu-Dodan and everyone who helped! I'm excited about this. :rocket:

toddbaert (approved):

service/openapi.yaml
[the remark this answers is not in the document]
Kavindu-Dodan: In the community meeting on 29.02.2024, we decided not to include any type details in either request and response

This decision is now documented in a dedicated ADR [1]. We could add these details if we see the requirement in a future iteration. 

If no objections, I will resolve this conversation before merging the PR. 

[1] - https://github.com/open-feature/protocol/blob/0896456b001f217c53a155d6075854ef58ff06b2/service/adrs/0003-no-flag-type-in-request-or-response.md

service/openapi.yaml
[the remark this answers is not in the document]
Kavindu-Dodan: In the community meeting on 29.02.2024, we decided not to include any type details in either request and response

This decision is now documented in a dedicated ADR [1]. We could add these details if we see the requirement in a future iteration. 

If no objections, I will resolve this conversation before merging the PR. 

[1] - https://github.com/open-feature/protocol/blob/0896456b001f217c53a155d6075854ef58ff06b2/service/adrs/0003-no-flag-type-in-request-or-response.md

nicklasl (approved):
🚀 Good stuff! Love the ADRs!

service/openapi.yaml
```diff
@@ -0,0 +1,241 @@
...
+          description: Rate limit reached on the Flag Management System
+          headers:
+            Retry-Later:
+              description: Indicates when to retry the request again
+              schema:
+                type: string
+                format: date-time
+                examples:
+                  - '2024-02-07T12:00:00Z'
+        '500':
```
dyladan, line 56: These errors have no schema and the default is just `"string"`. Should we include `#/components/schemas/evaluationFailure` in these responses? Same question for the other endpoint.

dyladan (approved):
Looks good to me. I like having the ADRs as well

service/openapi.yaml
```diff
@@ -0,0 +1,241 @@
...
+            schema:
+              $ref: '#/components/schemas/evaluationRequest'
+      responses:
+        '200':
+          description: OFREP successful evaluation response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/evaluationSuccess'
+        '400':
```
jonathannorris, line 36: Shouldn't there be a 404 response for a key that isn't found?

jonathannorris: Seeing that we do have an error code for `"errorCode": "FLAG_NOT_FOUND",` 

Or is the intention that we always return the default value with `"reason": "DEFAULT"`, its a bit ambiguous how providers should implement this.

jonathannorris: Just saw the discussion here that a 404 should be captured by a `400` error: https://github.com/open-feature/protocol/pull/2#issuecomment-1965120023

I disagree that we we should be capturing `404` errors with a more generic `400` error, we should not be trying to change standard HTTP error code behaviour, for most API consumers it will be very un-natural that a not-found error is a `400` not a `404`.

Kavindu-Dodan: We already have HTTP 400 to communicate back bad flag evaluation responses mapping to error codes [1] and this is why I removed the dedicated 404 response. And if the provider receives a 400 (or any other error), the evaluation should be `ERROR`.

Besides, in my view flag evaluation request is not necessarily a request to obtain a resource but rather an execution of an operation. So a 404 should mean `missing operation` which translates to a missing flag evaluation endpoint which we will have to handle in the provider implementation. 

 (cc - @lukas-reining  @thomaspoignant appreciate opinion on this )

[1] - https://openfeature.dev/specification/types#error-code

service/openapi.yaml
```diff
@@ -0,0 +1,241 @@
...
+            - $ref: "#/components/schemas/floatFlag"
+            - $ref: "#/components/schemas/objectFlag"
+    evaluationFailure:
+      description: Flag evaluation failure response
+      properties:
+        key:
+          $ref: '#/components/schemas/key'
+        errorCode:
+          type: string
+          enum: [ FLAG_NOT_FOUND,PARSE_ERROR,TYPE_MISMATCH,TARGETING_KEY_MISSING,INVALID_CONTEXT,GENERAL ]
```
jonathannorris, line 180: Does it make sense to still have `TYPE_MISMATCH` if we have removed types from the API?

service/openapi.yaml
Kavindu-Dodan: If we see the need for an error payload, for example for logging purposes, we could include one with at least an error message

service/openapi.yaml
Kavindu-Dodan: Good point, we could remove this and transfer the responsibility of generating this to the provider implementation.

service/openapi.yaml
thomaspoignant: I have no strong opinion on this one.
I found the 404 handy since it is explicit from an API perspective.

But I can live without it also.

beeme1mr: A 404 makes sense when attempting to evaluate an individual flag.

service/openapi.yaml
lukas-reining: >  we could remove this and transfer the responsibility of generating this to the provider implementation

Yes, I think we should do this, it does not make sense here.

service/openapi.yaml
lukas-reining: > I found the 404 handy since it is explicit from an API perspective.

Staying close to specific HTTP semantic generally feels like a good thing, so going for 404 for single evaluations seems good, on the other hand one might argue that all errors are expressed through the 400 with an error code in our case, which is still HTTP compliant and might give us a simpler interface. So to me, parsing 400 errors would always be necessary and 404 would not give us any additional information. 
Yes it might give us a shortcut to not have to look into the response body @thomaspoignant, still I would expect a `evaluationFailure` response here.

From the [RFC](https://httpwg.org/specs/rfc9110.html#status.401):
> A 404 response is heuristically cacheable; i.e., unless otherwise indicated by the method definition or explicit cache controls (see [Section 4.2.2](https://httpwg.org/specs/rfc9111.html#heuristic.freshness) of [[CACHING]](https://httpwg.org/specs/rfc9110.html#CACHING)).

I am not sure if this might be important to keep in mind. Maybe we do or do not want this to be possibly done by middle boxes like proxys? 

> So a 404 should could mean missing operation which translates to a missing flag evaluation endpoint which we will have to handle in the provider implementation.

By the specification a 404 for `/ofrep/v1/evaluate/flags/{key}` would always mean that the flag is not found if the service is OFREP compliant so I am not sure if that is really the case here @Kavindu-Dodan.


Generally I think both 400 and 404 would be okay. I am leaning towards a 404 with the same `evaluationFailure` response as for the other 400s.

service/openapi.yaml
Kavindu-Dodan: If no objections, I will define this and add to be used by appropriate 4xxx & 5xx responses

service/openapi.yaml
toddbaert: > am leaning towards a 404 with the same evaluationFailure response as for the other 400s.

Agreed.

Kavindu-Dodan: Thank you @jonathannorris for starting this conversation. And based on feedback from @thomaspoignant @beeme1mr @lukas-reining & @toddbaert  I will add a dedicated 404 response with error code set to `FLAG_NOT_FOUND `

service/openapi.yaml
```diff
@@ -0,0 +1,241 @@
...
+          $ref: '#/components/schemas/context'
+    evaluationSuccess:
+      description: Flag evaluation success response.
+      allOf:
+        - properties:
+            key:
+              $ref: "#/components/schemas/key"
+            reason:
+              type: string
+              description: Reason matching OpenFeature reason
```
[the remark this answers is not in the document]
Kavindu-Dodan: I would also include `STATIC` on this list

service/openapi.yaml
Kavindu-Dodan: This is complete with the commit 10eda958c989cfc9ad79b89d9b302ade684f0f5b

service/openapi.yaml
Kavindu-Dodan: Removed `TYPE_MISMATCH ` and committed with https://github.com/open-feature/protocol/commit/10eda958c989cfc9ad79b89d9b302ade684f0f5b

service/openapi.yaml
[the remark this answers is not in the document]
Kavindu-Dodan: I added the specific reasons to the response and they now contain enum with `STATIC,TARGETING_MATCH,SPLIT,DISABLED,UNKNOWN ` values.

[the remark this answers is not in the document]
jonathannorris: Why `STATIC`, what's an example of where it would apply for this case? 

definition: `The resolved value is static (no dynamic evaluation).`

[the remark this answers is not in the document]
Kavindu-Dodan: `STATIC` reason can occur if the value is derived as a non-dynamic evaluation from the flag service point of view. AS a fact flagd uses `STATIC` reason when flag evaluation is done without any targeting rules (involvement of context data).

jonathannorris (approved):

Kavindu-Dodan:
If there are no more concerns, I will merge this version of the OpenAPI specification on **07.03.2024**. 

Once merged, we could continue to improve the API through issues and follow-up PRs. I expect such changes to be non-breaking so that OFREP adaptions can proceed smoothly.

PR #7, opened 2024-02-21
Simplify OFREP OpenAPI Definition

beeme1mr:
## This PR

- Updated the URI from evaluate to flags. Semi-related to https://github.com/open-feature/protocol/pull/2#discussion_r1496331648
- Removed the list of flags from the bulk eval API. Relates to https://github.com/open-feature/protocol/pull/2#discussion_r1493008250
- Removed request type from the request. Relates to https://github.com/open-feature/protocol/pull/2#discussion_r1472797815
- Removed ETag support. Relates to https://github.com/open-feature/protocol/pull/2#discussion_r1493008250
- Remove change API due to redundant behavior. Relates to https://github.com/open-feature/protocol/pull/2#discussion_r1496323381
- Updated the bulk API response to be an object. This will allow us to extend the response in a nonbreaking way in the future.

### Notes

This is a simplified version of the current proposal. I believe, in this case, less is more. It tends to be much easier to add new functionality than to change something. In my opinion, we should defer ETag and caching behavior until after we have the basics in place. I believe this proposal will work well when using both static and dynamic paradigms.

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.1.0
 info:
   version: 0.1.0
   title: OpenFeature Remote Evaluation Protocol (OFREP)
@@ -11,9 +11,11 @@ security:
   - ApiKeyAuth: []
   - BearerAuth: []
 paths:
-  /ofrep/v1/evaluate/{key}:
+  /ofrep/v1/flags/{key}:
     post:
-      description: OFREP flag evaluation request
+      description: Evaluate an individual flag
+      tags:
+        - evaluate
       parameters:
         - name: key
           in: path
@@ -59,9 +61,11 @@ paths:
                 example: '2024-02-07T12:00:00Z'
         '500':
           description: Internal server error
-  /ofrep/v1/evaluate:
+  /ofrep/v1/flags:
     post:
-      description: OFREP bulk evaluation request
+      description: Bulk evaluate multiple flags
+      tags:
+        - evaluate
       requestBody:
         content:
           application/json:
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -11,9 +11,11 @@ security:
   - ApiKeyAuth: []
   - BearerAuth: []
 paths:
-  /ofrep/v1/evaluate/{key}:
+  /ofrep/v1/flags/{key}:
```
thomaspoignant, line 14: The proposal in the comment was to use `/ofrep/v1/evaluate/flags.`
I've already edited the other branch with the proposal.

beeme1mr: I'm fine with that too 👍

jonathannorris: How necessary do we think `/ofrep/` is?

service/openapi.yaml
```diff
@@ -95,103 +99,58 @@ paths:
...
-          description: Error code specific to the bulk evaluation.
-        errorDetails:
-          type: string
-          description: Optional error details description for logging or other needs
-      required:
-        - errorCode
-    flagChangeFailure:
-      description: Bulk evaluation failure response
+        flags:
+          type: array
```
jonathannorris, line 123: Making the top level here an array will be pretty limiting to us in the future. If we ever wanted to allow the providers to respond with any further configuration in the future (for example: polling rate info / sse connection details / event configuration) it may be wise to have the top level here be an object:

```
{ "flags": [{...}] }
```

service/openapi.yaml
```diff
@@ -204,8 +163,6 @@ components:
     evaluationRequest:
       description: Flag evaluation request
       properties:
-        flagType:
-          $ref: '#/components/schemas/flagType'
         context:
           $ref: '#/components/schemas/context'
     success:
```
jonathannorris, line 168: Rename to `evaluationSuccess` to be consistent with `bulkEvaluationSuccess`. 

Also, should we just include the same `errorCode` and `errorDetails` params here and just make the `flags` in `bulkEvaluationSuccess` use this shared schema?

jonathannorris: also rename the `failure` schema

service/openapi.yaml
beeme1mr: > Actually, I think I am reading the OpenAPI spec wrong here; it is a top-level object.

Yeah, exactly. This PR updates the response to be an object instead of an array the the reasons you mentioned in your first comment.

service/openapi.yaml
thomaspoignant: You have a discussion about that in the other PR.
We added /ofrep/ to avoid any collision with existing v1 that can exist in other systems

service/openapi.yaml
thomaspoignant: I’ve actually made that change in the other PR

PR #9, opened 2024-04-07
feat: Add metadata endpoint

thomaspoignant:
## This PR
This PR adds the support of a new endpoint  `/ofrep/v1/configuration` in the OFREP OpenAPI spec.  
This new endpoint is here to be able to expose some information about the flag management system to the generic provider.

This endpoint will be called during the [initialization](https://openfeature.dev/specification/sections/providers/#24-initialization) of the provider and will allow the provider to configure itself based on the configuration received.

This endpoint will return basic information such as the `name` of the flag management system. 
But also what `features` are supported by the remote flag management systems,  and basic remote configuration on how the provider should behave.

This 1st version does not contain a lot of possible configuration, but we will be able to extend it as we continue to evolve the OpenFeature Remote Evaluation Protocol.

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -34,7 +34,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/configuraitonResponse'
+                $ref: '#/components/schemas/configurationResponse'
         '304':
           description: Flag Management System Metadata is not modified
         '401':
@@ -314,7 +314,7 @@ components:
     errorDetails:
       type: string
       description: An error description for logging or other needs
-    configuraitonResponse:
+    configurationResponse:
       description: OFREP metadata response
       properties:
         name:
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -12,10 +12,10 @@ security:
   - ApiKeyAuth: [ ]
   - BearerAuth: [ ]
 paths:
-  /ofrep/v1/metadata:
+  /ofrep/v1/configuration:
     get:
-      summary: OFREP metadata
-      description: OFREP metadata to provide information about the remote flag management system, to configure the OpenFeature SDK providers. This endpoint is called during the initialization of the provider.
+      summary: OFREP provider configuration
+      description: OFREP configuration to provide information about the remote flag management system, to configure the OpenFeature SDK providers. This endpoint will be called during the initialization of the provider.
       parameters:
         - in: header
           name: If-None-Match
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -16,13 +16,27 @@ paths:
     get:
       summary: OFREP metadata
       description: OFREP metadata , it gives information about the remote flag management system, to configure the providers. This endpoint is called during the initialization of the provider.
+      parameters:
+        - in: header
+          name: If-None-Match
+          description: The request will be processed only if ETag doesn't match any of the values listed.
+          schema:
+            type: string
+          required: false
       responses:
         '200':
           description: OFREP metadata response
+          headers:
+            ETag:
+              schema:
+                type: string
+              description: Entity tag used for cache validation
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/metadataResponse'
+        '304':
+          description: Flag Management System Metadata is not modified
         '401':
           description: Unauthorized - You need credentials to access the API
         '403':
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -344,7 +344,7 @@ components:
           description: set to true if the remote flag management system is supporting polling
         minPollingInterval:
           type: number
-          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to  set any polling value under this minimum.
+          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
           examples:
             - 60000
       required:
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -325,7 +325,7 @@ components:
             - go-feature-flag
         capabilities:
           type: object
-          description: List of the features supported by the flag management system.
+          description: Capabilities of the flag management system and how to configure them in the provider.
           properties:
             cacheInvalidation:
               $ref: '#/components/schemas/featureCacheInvalidation'
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -323,7 +323,7 @@ components:
           examples:
             - flagd
             - go-feature-flag
-        features:
+        capabilities:
           type: object
           description: List of the features supported by the flag management system.
           properties:
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -342,12 +342,10 @@ components:
         enabled:
           type: boolean
           description: set to true if the remote flag management system is supporting polling
-          default: false
         minPollingInterval:
           type: number
           description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
           examples:
             - 60000
-          default: 60000
       required:
         - name
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -342,10 +342,12 @@ components:
         enabled:
           type: boolean
           description: set to true if the remote flag management system is supporting polling
+          default: false
         minPollingInterval:
           type: number
           description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
           examples:
             - 60000
+          default: 60000
       required:
         - name
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -15,7 +15,7 @@ paths:
   /ofrep/v1/metadata:
     get:
       summary: OFREP metadata
-      description: OFREP metadata , it gives information about the remote flag management system, to configure the providers. This endpoint is called during the initialization of the provider.
+      description: OFREP metadata to provide information about the remote flag management system, to configure the OpenFeature SDK providers. This endpoint is called during the initialization of the provider.
       parameters:
         - in: header
           name: If-None-Match
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -12,6 +12,27 @@ security:
   - ApiKeyAuth: [ ]
   - BearerAuth: [ ]
 paths:
+  /ofrep/v1/metadata:
+    get:
+      summary: OFREP metadata
+      description: OFREP metadata , it gives information about the remote flag management system, to configure the providers. This endpoint is called during the initialization of the provider.
+      responses:
+        '200':
+          description: OFREP metadata response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/metadataResponse'
+        '401':
+          description: Unauthorized - You need credentials to access the API
+        '403':
+          description: Forbidden - You are not authorized to access the API
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/generalErrorResponse'
   /ofrep/v1/evaluate/flags/{key}:
     post:
       description: OFREP single flag evaluation request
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -12,6 +12,27 @@ security:
   - ApiKeyAuth: [ ]
   - BearerAuth: [ ]
 paths:
+  /ofrep/v1/metadata:
```
nicklasl, line 15: This type of endpoint is an excellent example on when caching can be used. Let's have that included as well 👍

thomaspoignant: Since it will called only during initialization that's why I had not included any caching.

What did you had in mind? The same mechanism that we have for the ` /ofrep/v1/evaluate/flags` endpoint _(with a 304 response + a management of the `If-none-match`)_?

nicklasl: Yes exactly like the others, etag + if-none-match.
Maybe it doesn't make much sense in "volatile" environments like JS, but for mobile environments the http caching is often persisted on disk and then it would be good to have.

thomaspoignant: I've added the support of etag + if-none-match.

service/openapi.yaml
```diff
@@ -279,3 +314,33 @@ components:
...
+          examples:
+            - flagd
+            - go-feature-flag
+        features:
+          type: object
+          description: List of the features supported by the flag management system.
+          properties:
+            cacheInvalidation:
+              type: array
+              enum: [POLLING]
```
nicklasl, line 332: Each of these `cacheInvalidation` types would need its own documentation. Where do we keep those?

service/openapi.yaml
```diff
@@ -279,3 +314,33 @@ components:
...
+    metadataResponse:
+      description: OFREP metadata response
+      properties:
+        name:
+          type: string
+          description: name of the flag management system
+          examples:
+            - flagd
+            - go-feature-flag
+        features:
```
nicklasl, line 326: `minPollingInterval` from `configurations` below is clearly connected with one of these "features"? Is it feasibile to come up with a format to pass the `configuration` as part of the `feature`?

thomaspoignant: Would something like this example look better to you?

```json
{
   "name": "flagD",
   "features": {
      "cacheInvalidation": {
         "supported": ["POLLING"],
         "polling": {
            "minPollingInterval": 1000
         }
      }
   }
}
```

service/openapi.yaml
thomaspoignant: For now we only have the support of 1 type of cache invalidation (aka polling), this is more here to extendable in the future.

service/openapi.yaml
thomaspoignant: I like the first one, supporting the 2 syntax with the same key will be not super easy in some languages that are not supporting combine types.

I can update the openAPI specs with the 1st proposition if you're ok.

nicklasl: yeah, I like the first one as well 👍

service/openapi.yaml
```diff
@@ -12,6 +12,41 @@ security:
   - ApiKeyAuth: [ ]
   - BearerAuth: [ ]
 paths:
+  /ofrep/v1/metadata:
+    get:
+      summary: OFREP metadata
+      description: OFREP metadata , it gives information about the remote flag management system, to configure the providers. This endpoint is called during the initialization of the provider.
```
jonathannorris, line 18:
```suggestion
      description: OFREP metadata to provide information about the remote flag management system, to configure the OpenFeature SDK providers. This endpoint is called during the initialization of the provider.
```

service/openapi.yaml
jonathannorris: Not a fan of using `features` here, as most providers call themselves `Feature Management Platforms` now, and many systems have the concept of `features` as a top-level objects that will cause confusion here. I'd stick to calling this `configuration` / `options` something in that line.

service/openapi.yaml
```diff
@@ -279,3 +314,38 @@ components:
...
+    featureCacheInvalidationPolling:
+      type: object
+      description: Configuration of the polling for the featureCacheInvalidation
+      properties:
+        enabled:
+          type: boolean
+          description: set to true if the remote flag management system is supporting polling
+        minPollingInterval:
+          type: number
+          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to  set any polling value under this minimum.
```
kinyoklion, line 347:
```suggestion
          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
```

service/openapi.yaml
```diff
@@ -279,3 +314,38 @@ components:
...
+      description: Configuration of the polling for the featureCacheInvalidation
+      properties:
+        enabled:
+          type: boolean
+          description: set to true if the remote flag management system is supporting polling
+        minPollingInterval:
+          type: number
+          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
+          examples:
+            - 60000
```
beeme1mr, line 349: Where should we specify the defaults? In this case, I would expect polling to be disabled if this is omitted but it would be nice to explicitly state that behavior somewhere.

kinyoklion: I think I need a better explanation of this feature specifically.

Like if a provider is using a vendor SDK configured with streaming, does that mean it doesn't support polling? Or is the goal here cache freshness value, in which case it would just be however recent you want your flag cache to be. (Correlated to polling because requesting flags greatly faster than the configured polling interval is wasteful.)

weyert: Wouldn't this be more `configuration` instead of `metadata`?

lukas-reining: > Nothing prevents vendors to create their own providers and use streaming instead of polling but this will not be offered out of the box by the OFREP providers.

Well not yet right @thomaspoignant? It is really possible that we are doing some server push like SSE in the future, just not yet. And sure, the config only applies to the OFREP providers.

> > Wouldn't this be more configuration instead of metadata?
>
>@weyert Yes, it could be a better name, I wonder what other things about the name?

I also think configuration would be better suited than metadata. The only thing is the name, which might be good for the metadata of the evaluation details, but we could also just decide that this is mapped to them then.

thomaspoignant: >Well not yet right @thomaspoignant? It is really possible that we are doing some server push like SSE in the future, just not yet. And sure, the config only applies to the OFREP providers.

Yes, it will probably come, SSE or even websockets, but this is not part of this 1st version.

>I also think configuration would be better suited than metadata. The only thing is the name, which might be good for the metadata of the evaluation details, but we could also just decide that this is mapped to them then.

I'll change the endpoint to `/ofrep/v1/configuration`.

service/openapi.yaml
```diff
@@ -279,3 +314,40 @@ components:
...
+    metadataResponse:
+      description: OFREP metadata response
+      properties:
+        name:
+          type: string
+          description: name of the flag management system
+          examples:
+            - flagd
+            - go-feature-flag
+        features:
```
lukas-reining, line 326: As @jonathannorris said

> Not a fan of using features here, as most providers call themselves Feature Management Platforms now, and many systems have the concept of features as a top-level objects that will cause confusion here. I'd stick to calling this configuration / options something in that line.

I would also prefer something like configuration here.
Feature is cool for this case but there might be an option in the future that is not directly bound to something you would call a feature. But as I am struggling to come up with an example it might also be fine to go with features.

kinyoklion: "capabilities" is a good fallback when you want "features" in your feature management. Assuming that they are features/capabilities "I support the polling capability."

lukas-reining: Thats also nice, I am still just not sure if there will be configurations that are not `capabilities` @kinyoklion.
But as said that is just feeling I am still struggling to come up with an example.

service/openapi.yaml
```diff
@@ -279,3 +314,40 @@ components:
...
+        enabled:
+          type: boolean
+          description: set to true if the remote flag management system is supporting polling
+          default: false
+        minPollingInterval:
+          type: number
+          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
+          examples:
+            - 60000
+          default: 60000
```
lukas-reining, line 351: I am not sure about having a default value here.
For a protocol implemented by different vendors and projects, I think having default for timing that are not technically constrained is not too helpful.
Is there any specific reason for this value @thomaspoignant?

service/openapi.yaml
thomaspoignant: I like capabilities because this is really the goal.

And if we have other configurations (i don't think of any use case now) we can always add another key in the API.

service/openapi.yaml
thomaspoignant: It was to be explicit on what the provider should implement as a default value.
It is also based on @beeme1mr comment https://github.com/open-feature/protocol/pull/9#discussion_r1556377284.

service/openapi.yaml
```diff
@@ -279,3 +314,40 @@ components:
     errorDetails:
       type: string
       description: An error description for logging or other needs
+    configuraitonResponse:
```
nicklasl, line 317:
```suggestion
    configurationResponse:
```

service/openapi.yaml
```diff
@@ -12,6 +12,41 @@ security:
...
+          description: OFREP metadata response
+          headers:
+            ETag:
+              schema:
+                type: string
+              description: Entity tag used for cache validation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/configuraitonResponse'
```
nicklasl, line 37:
```suggestion
                $ref: '#/components/schemas/configurationResponse'
```

service/openapi.yaml
lukas-reining: I see, did not see that this was the motivation for the default.
As I understood @beeme1mr he would not expect a default but rather a behavior if it is omitted.
I would also not add a default value for this specific parameter, as I think it just depends in this case.

thomaspoignant: I don't have a strong feeling to keep it or not.
We can remove it if you feel that it is needed.

For the omitted behaviour, I am not sure if OpenAPI has a way to represent that?

lukas-reining: Mh I think we can only describe the omitted behavior.
I am more against having a default value for the interval as it is so dependent from the concrete system and use case, and I don't see a way of finding a "good" default the one could argue about.

beeme1mr: I'm in favor of removing the defaults but I think we'll need a separate document describing the expected behavior of the providers.

service/openapi.yaml
```diff
@@ -279,3 +314,40 @@ components:
...
+      properties:
+        name:
+          type: string
+          description: name of the flag management system
+          examples:
+            - flagd
+            - go-feature-flag
+        capabilities:
+          type: object
+          description: List of the features supported by the flag management system.
```
beeme1mr, line 328: The description no longer matches the implementation.

service/openapi.yaml
beeme1mr: For example, we need to define what happens in the provider if a user defines a polling interval faster than the flag management tools allow.

service/openapi.yaml
```diff
@@ -279,3 +314,40 @@ components:
...
+      type: object
+      description: Configuration of the polling for the featureCacheInvalidation
+      properties:
+        enabled:
+          type: boolean
+          description: set to true if the remote flag management system is supporting polling
+          default: false
+        minPollingInterval:
+          type: number
+          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
```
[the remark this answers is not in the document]
thomaspoignant: Yes, I think it is already a bit late to change that in the provider.

service/openapi.yaml
thomaspoignant: > I'm in favor of removing the defaults but I think we'll need a separate document describing the expected behavior of the providers.

This may be something we add in the ADRs?
Would it work for you?

beeme1mr: It's more of an OFREP provider spec than an architecture decision. I was thinking about something like this 
https://flagd.dev/reference/specifications/in-process-providers/

lukas-reining: > It's more of an OFREP provider spec than an architecture decision.

Yes I would say so too.
I think it should be a provider spec, as OFREP and the OpenAPI Docs are somewhat disconnected from that.

thomaspoignant: Yes, that was the initial plan we had with @Kavindu-Dodan.
I'll try to have a 1st draft tonight.

beeme1mr: Awesome, thanks!

beeme1mr (approved):

lukas-reining (approved):

service/openapi.yaml
Kavindu-Dodan: +1 for defining default values in the spec. 

Besides, we can make the metadata endpoint optional as the spec will define the defaults.

PR #12, opened 2024-04-23
feat: Add unsupported evaluation types

thomaspoignant:
## This PR
In this PR we add a new configuration in the `/ofrep/v1/configuration` endpoint that lets the flag management system specify which types the system is not supporting.

By default, we consider that all flag types are supported, but if a type is specified in `unsupportedTypes` we will return an error every time someone is trying to call the associated evaluation function.

### Example
If the configuration endpoint is returning something like this:
```json
{
  // ...
  "capabilities": {
	"evaluation": {
	  "unsupportedTypes": ["object", "int", "float"]
    }
  }
}
```

When using the provider, and calling the `ObjectValue` we expect to get an error without trying to call the remote flag management system.

This function will always return an error and the default value without doing any HTTP calls.
```go
objectValue, err := client.ObjectValue(context.Background(), "objectFlag", myObject, openfeature.EvaluationContext{})
```

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -330,8 +330,8 @@ components:
             cacheInvalidation:
               $ref: '#/components/schemas/featureCacheInvalidation'
             flagEvaluation:
-              $ref: '#/components/schemas/featureEvaluation'
-    featureEvaluation:
+              $ref: '#/components/schemas/flagEvaluation'
+    flagEvaluation:
       type: object
       description: Configurations specific for flag evaluations in OFREP provider implementation
       properties:
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -336,7 +336,7 @@ components:
       description: Configuration on how to do flag evaluation in the providers.
       properties:
         unsupportedTypes:
-          description: List of unsupported types by the flag management system, it will return an error if we call the type associated function in the OFREP providers.
+          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value
           type: array
           items:
             type: string
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -333,7 +333,7 @@ components:
               $ref: '#/components/schemas/featureEvaluation'
     featureEvaluation:
       type: object
-      description: Configurations specific for flag evaluations at OFREP provider implementation
+      description: Configurations specific for flag evaluations in OFREP provider implementation
       properties:
         unsupportedTypes:
           description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -336,7 +336,7 @@ components:
       description: Configurations specific for flag evaluations in OFREP provider implementation
       properties:
         unsupportedTypes:
-          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -329,6 +329,19 @@ components:
           properties:
             cacheInvalidation:
               $ref: '#/components/schemas/featureCacheInvalidation'
+            evaluation:
+              $ref: '#/components/schemas/featureEvaluation'
+    featureEvaluation:
+      type: object
+      description: Configuration for the cache cacheInvalidation
```
markphelps, line 336: this seems like a copy paste of the description below, is this correct?

thomaspoignant: oh good catch, let me change that

service/openapi.yaml
```diff
@@ -329,6 +329,20 @@ components:
           properties:
             cacheInvalidation:
               $ref: '#/components/schemas/featureCacheInvalidation'
+            evaluation:
+              $ref: '#/components/schemas/featureEvaluation'
+    featureEvaluation:
```
Kavindu-Dodan, line 334: suggesting to rename to flagEvaluation, mapping the OF spec [1]

[1] - https://openfeature.dev/specification/sections/flag-evaluation#13-flag-evaluation

service/openapi.yaml
```diff
@@ -329,6 +329,20 @@ components:
...
+      type: object
+      description: Configuration on how to do flag evaluation in the providers.
```
Kavindu-Dodan, line 336:
```suggestion
      description: Configurations for flag evaluation at OFREP provider implementation
```

service/openapi.yaml
```diff
@@ -329,6 +329,20 @@ components:
...
+      properties:
+        unsupportedTypes:
+          description: List of unsupported types by the flag management system, it will return an error if we call the type associated function in the OFREP providers.
```
Kavindu-Dodan, line 339:
```suggestion
          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value
```

service/openapi.yaml
```diff
@@ -329,6 +329,20 @@ components:
           properties:
             cacheInvalidation:
               $ref: '#/components/schemas/featureCacheInvalidation'
+            flagEvaluation:
+              $ref: '#/components/schemas/featureEvaluation'
+    featureEvaluation:
+      type: object
+      description: Configurations specific for flag evaluations at OFREP provider implementation
```
lukas-reining, line 336:
```suggestion
      description: Configurations specific for flag evaluations in OFREP provider implementation
```

Not 100% sure.

lukas-reining (approved):
Looks good to me!

service/openapi.yaml
```diff
@@ -329,6 +329,20 @@ components:
...
             cacheInvalidation:
               $ref: '#/components/schemas/featureCacheInvalidation'
+            flagEvaluation:
+              $ref: '#/components/schemas/featureEvaluation'
+    featureEvaluation:
+      type: object
+      description: Configurations specific for flag evaluations in OFREP provider implementation
+      properties:
+        unsupportedTypes:
+          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value
```
beeme1mr, line 339:
```suggestion
          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value.
```

service/openapi.yaml
beeme1mr: Please also update the ref to use the updated name.

beeme1mr (approved):

PR #13, opened 2024-04-30
chore: add linter for OpenAPI

thomaspoignant:
Add linter for OpenAPI validation.

### DIFF
--- a/.github/workflows/lint-openapi.yaml
+++ b/.github/workflows/lint-openapi.yaml
@@ -1,0 +1,23 @@
+name: "OpenAPI Lint"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  openAPILint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: OpenAPI Lint Checks
+        uses: nwestfall/openapi-action@v1.0.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          file: services/openapi.yaml
--- a/.github/workflows/lint-openapi.yaml
+++ b/.github/workflows/lint-openapi.yaml
@@ -7,9 +7,6 @@ on:
       - edited
       - synchronize
 
-permissions:
-  contents: read
-
 jobs:
   openAPILint:
     runs-on: ubuntu-latest


.,
Kavindu-Dodan, line 1: I think this is not needed

thomaspoignant: definitely not, I am not sure why it is here 🤦

.github/workflows/lint-openapi.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - synchronize
+
+jobs:
+  openAPILint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: OpenAPI Lint Checks
+        uses: nwestfall/openapi-action@v1.0.2
```
Kavindu-Dodan, line 17: Regarding this action, it seems this has been unmaintained for some time. And I checked alternatives and came across redocly-cli (MIT license) and used their docker image [1] to perform a similar validation 

You can see my attempt in the fork [2] and the test GH action failure run in my fork [3]

Let me know your thoughts on this :) 

[1] - https://github.com/Redocly/redocly-cli?tab=readme-ov-file#docker
[2] - https://github.com/open-feature/protocol/compare/test-validate
[3] - https://github.com/open-feature/protocol/actions/runs/8913421964/job/24478855581

thomaspoignant:
@Kavindu-Dodan Maybe we should take your branch instead of this one.
When it comes to the validation I am not sure why it fails, but this is worth checking.

.github/workflows/lint-openapi.yaml
beeme1mr: This is another option that adds lint violations as annotations in the PR.

https://github.com/stoplightio/spectral-action

Kavindu-Dodan:
> @Kavindu-Dodan Maybe we should take your branch instead of this one. When it comes to the validation I am not sure why it fails, but this is worth checking.

I will take a shot on this then :)

Kavindu-Dodan:
@thomaspoignant here is a PR with redocly - https://github.com/open-feature/protocol/pull/15

.github/workflows/lint-openapi.yaml
Kavindu-Dodan: @beeme1mr checked but I am not super happy yet as there are open issues and PRs that were not addressed recently. We can switch back if we see issues with redocly

thomaspoignant:
Close this PR in favor of #15

Issue #11, opened 2024-04-19
Create provider guidelines


thomaspoignant:
# Description
Writing an OFREP provider requires some specificity.
We should create a provider guidelines to explain how to interact with the API.

PR #14, opened 2024-04-30
feat: Client provider guidelines

thomaspoignant:
This PR is a first draft for a provider client spec.

Please provide feedback on this one to know if I am heading in the right direction and to know if it can be helpful to create providers.

Those are the guidelines related to the issue #11


provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
+# Creating an OFREP client provider
+
+OpenFeature Remote Evaluation Protocol (OFREP) is an API specification for feature flagging that allows the use of generic providers to connect to any feature flag management systems that supports the protocol.
+
+In this specification we will specify how to write a OFREP provider using the [static-context-paradigm](https://openfeature.dev/specification/glossary/#static-context-paradigm) that is used on client side applications typically operate in the context of a single user.
```
markphelps, line 5:
```suggestion
In this specification we will specify how to write an OFREP provider using the [static-context-paradigm](https://openfeature.dev/specification/glossary/#static-context-paradigm) that is used on client side applications typically operate in the context of a single user. 
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+We will keep the specification language agnostic.
+
+**Pre-requisite:**
+- Understanding of [general provider concepts](https://openfeature.dev/docs/reference/concepts/provider/)
+- Understanding of the [OFREP](../../README.md)
+- Understanding of the [OFREP OpenAPI specification](../../service/openapi.yaml)
+
+
+## Constructor
+An implementation of a OFREP client provider should allow in the creation of the provider to take as options:
```
markphelps, line 15:
```suggestion
An implementation of an OFREP client provider should allow in the creation of the provider to take as options:
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+- `baseURL`: The base URL of the [flag management system](https://openfeature.dev/specification/glossary#flag-management-system). This should be the base of the URL pointing before the `/ofrep` namespace of the API.
+- `headers`: The headers to use when calling the OFREP endpoints *(ex:`Authorization`, Custom headers, etc ...)*.
+- `pollInteral`: The polling interval defining when to call again the flag management system.
+
+In the constructor the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
+
+## Initialize the provider
+An implementation of a OFREP client provider should start with an initialization of the provider.
```
markphelps, line 23:
```suggestion
An implementation of an OFREP client provider should start with an initialization of the provider.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+
+The `initialize()` function should follow those steps:
+1. Make a GET request to the `/ofrep/v1/configuration` endpoint to retrieve the configurations return by the flag management system and store them in memory to be available for all the function of the provider. *(See [Annexe 1](#annexe-1) for the description of the endpoint response)*
+2. Make a POST request to the `/ofrep/v1/evaluate/flags` endpoint with the evaluation context in the body.  
+   - If the endpoint return an error, the `initialize()` function must error and exit.
```
markphelps, line 28:
```suggestion
   - If the endpoint returns an error, the `initialize()` function must error and exit.  
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+In the constructor the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
+
+## Initialize the provider
+An implementation of a OFREP client provider should start with an initialization of the provider.
+
+The `initialize()` function should follow those steps:
+1. Make a GET request to the `/ofrep/v1/configuration` endpoint to retrieve the configurations return by the flag management system and store them in memory to be available for all the function of the provider. *(See [Annexe 1](#annexe-1) for the description of the endpoint response)*
+2. Make a POST request to the `/ofrep/v1/evaluate/flags` endpoint with the evaluation context in the body.  
+   - If the endpoint return an error, the `initialize()` function must error and exit.  
+   - If the endpoint is in success, we should store in a local cache all the flags evaluation result returned by the API in a local cache. We should also store the `ETag` header in the provider to be able to send it back later.
```
markphelps, line 29:
```suggestion
   - If the request is successful, we should store in a local cache all of the flags evaluation results returned by the API in a local cache. We should also store the `ETag` header in the provider to be able to send it back later.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+3. If polling is enabled, the function should start the polling loop *(See [polling section](#polling))*.
+
+## Evaluation
+The evaluation should not do any remote API calls.
```
markphelps, line 33:
```suggestion
The evaluation should not perform any remote API calls.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+If we have a `ETag` available we should always add the header `If-None-Match` with the `ETag` value.
+- If the cache is still up-to-date we will receive a `304` telling us that the nothing has changed on the flag management system side.
+- If the cache is outdated we will receive a `200` with the new values of all the flags. In that situation we should:
+  1. Replace the actual local cache of flags evaluations.
+  2. Store the new `ETag` value for the future call.
+
+
+## Annexe 1
+**`/ofrep/v1/configuration` description:**
+The endpoint will return a list of configuration for the provider:
```
markphelps, line 55:
```suggestion
The endpoint will return a list of configurations for the provider:
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+- `name`: Name of the flag management system, it should be used in the `metadata.name` name (ex: `OFREP Web Provider ${metadata.name}`).
+- `capabilities`: List of capabilities of the flag management system and their associated configuration. *Check the [capabilities section](#capabilities) for the details of each capability.*
+
+### Capabilities
+In this section we will describe each capabilities and describe their default value and usage.
```
markphelps, line 60:
```suggestion
In this section we will describe each capability and describe their default value and usage.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+
+#### Cache Invalidation
+The capability `cacheInvalidation` describe how the mechanism of cache invalidation is working.
```
markphelps, line 63:
```suggestion
The capability `cacheInvalidation` describes how the mechanism of cache invalidation works.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+If the key `polling` is not available, the provider should use those default values:
+```json
+{
+    "enabled": true,
+    "minPollingInterval": 0 
+}
+```
+
+#### Flag Evaluation
+`flagEvaluation`: define the how to managed flag evaluation in the provider.
```
markphelps, line 80:
```suggestion
`flagEvaluation`: define how to manage flag evaluation in the provider.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+- `unsupportedTypes`: Some flag management system are not supporting all types. This array should contains all the types not supported by the flag management system.Acceptable values are `int`, `float`,`string`, `boolean`,`object`.
```
markphelps, line 81:
```suggestion
- `unsupportedTypes`: Some flag management systems do not support all types. This array should contain all the types not supported by the flag management system. Acceptable values are `int`, `float`, `string`, `boolean`, and `object`.  
```

markphelps (commented):
couple minor typos/grammar suggestions

provider/specs/client.md
beeme1mr, line 5:
```suggestion
In this document, we will specify how to write an OFREP provider using the [static-context-paradigm](https://openfeature.dev/specification/glossary/#static-context-paradigm) that is used on client-side applications typically operate in the context of a single user. 
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+- Understanding of the [OFREP OpenAPI specification](../../service/openapi.yaml)
+
+
+## Constructor
+An implementation of a OFREP client provider should allow in the creation of the provider to take as options:
+- `baseURL`: The base URL of the [flag management system](https://openfeature.dev/specification/glossary#flag-management-system). This should be the base of the URL pointing before the `/ofrep` namespace of the API.
+- `headers`: The headers to use when calling the OFREP endpoints *(ex:`Authorization`, Custom headers, etc ...)*.
+- `pollInteral`: The polling interval defining when to call again the flag management system.
+
+In the constructor the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
```
beeme1mr, line 20:
```suggestion
In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+The evaluation should not do any remote API calls.
+
+When calling an evaluation function the provider should check if the associated type is supported, by checking the key `capabilities.flagEvaluation.unsupportedTypes` from the configuration endpoint.
+- If the type is unsupported, we should exit early and directly return an error.
+- If the type is supported, we should retrieve the flag from the local cache of the flags evaluation.
+  - If you are not able to find the flag, you should return an FlagNotFound error.
+  - If the remote evaluation for this flag has return an error, you should map the error in provider and return the associated error.
+  - If the value retrieve from the cache has a different type than the one expected you should return a TypeMismatch error.
+  - If the cached evaluation is in success you should return the evaluation response.
+
```
beeme1mr, line 42: Visualizing this flow in Mermaid would be helpful.

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+## Polling
+The polling system will make a POST request periodically tp the `/ofrep/v1/evaluate/flags` endpoint to check if there is a change in the flags evaluation to be able to store it.
```
beeme1mr, line 44:
```suggestion
The polling system will make a POST request periodically to the `/ofrep/v1/evaluate/flags` endpoint to check if there is a change in the flags evaluation to be able to store it.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+
+If we have a `ETag` available we should always add the header `If-None-Match` with the `ETag` value.
```
beeme1mr, line 46:
```suggestion
If an `ETag` is available we should always add the header `If-None-Match` with the `ETag` value.
```

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+  1. Replace the actual local cache of flags evaluations.
+  2. Store the new `ETag` value for the future call.
+
+
+## Annexe 1
+**`/ofrep/v1/configuration` description:**
+The endpoint will return a list of configuration for the provider:
+- `name`: Name of the flag management system, it should be used in the `metadata.name` name (ex: `OFREP Web Provider ${metadata.name}`).
+- `capabilities`: List of capabilities of the flag management system and their associated configuration. *Check the [capabilities section](#capabilities) for the details of each capability.*
+
```
beeme1mr, line 58: What's the expected behavior if the call to `/ofrep/v1/configuration` fails?

beeme1mr (commented):
Great start, thanks! I added a few suggestions but I think this will be helpful.

We may want to avoid calling it a spec (at least for now). I would recommend calling it a guideline.

provider/specs/client.md
```diff
@@ -0,0 +1,82 @@
...
+- `enabled`: if `true` the provider should poll the `/ofrep/v1/evaluate/flags` regularly to check if any flag evaluation has changed in the flag management system.
+- `minPollingInterval`: define the minimum polling interval acceptable by the flag management system. If for any reason the `pollInteral` provided in the constructor is lower than this `minPollingInterval` we should default on this value.
+
+If the key `polling` is not available, the provider should use those default values:
+```json
+{
+    "enabled": true,
+    "minPollingInterval": 0 
+}
```
Kavindu-Dodan, line 76: This is best represented as a table. Besides, I think we should define a non-zero `minPollingInterval` default value

Kavindu-Dodan (commented):
Thanks @thomaspoignant for getting started on this 🙌 

I have following general remarks, 

- Language - Given this is a spec and to be similar to OF Spec, I think we should generilize langauge to avoid "we" usage (ex:- `In this specification we will specify..` to `This specificaiton specify..`)
- Spec name & title - Given this is specific to client side (aka static context), a good name for file could be `client-provider`/ `static-context-provider`


Otherwise I think this is a great first step

provider/specs/client.md
Kavindu-Dodan: Should fallback to the defaults and I think we should define the default behaviour in this spec

provider/specs/client.md
thomaspoignant: What would be an acceptable value for you? I was thinking at something like 60 seconds.

provider/specs/client.md
thomaspoignant: Hum good question, we were wondering with @lukas-reining and I think we should put the provider in error and retry later (probably with an exponential backoff).

WDYT?

thomaspoignant:
> Thanks @thomaspoignant for getting started on this 🙌
> 
> I have following general remarks,
> 
>     * Language - Given this is a spec and to be similar to OF Spec, I think we should generilize langauge to avoid "we" usage (ex:- `In this specification we will specify..` to `This specificaiton specify..`)
> 
>     * Spec name & title - Given this is specific to client side (aka static context), a good name for file could be `client-provider`/ `static-context-provider`
> 
> 
> Otherwise I think this is a great first step

I have renamed the file and stopped using the term specification to avoid any confusion here.
Do you think we should continue to update the language if it is only a guideline?

Kavindu-Dodan:
> I have renamed the file and stopped using the term specification to avoid any confusion here.
> Do you think we should continue to update the language if it is only a guideline?

Thanks :) 

>  Do you think we should continue to update the language if it is only a guideline?

Yeah, can start simple and then convert this to a spec. So let's not spend too much time on this.

Kavindu-Dodan (approved):
I am happy with this so we can get started :)

provider/specs/client.md
thomaspoignant: If we want to configure it, I will be more in favour of having an opt-out rather than an opt-in.

The reasoning is that this will encourage vendors to provide the information on how it works but also to avoid using the default values, the information of this endpoint is here to help to work better with the flag management system, not using them by default will be a bit weird from my point of view.

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+- `headers`: The headers to use when calling the OFREP endpoints *(ex:`Authorization`, Custom headers, etc ...)*.
+- `pollInteral`: The polling interval defining when to call again the flag management system.
+
+In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
+
+## Initialize the provider
+An implementation of an OFREP client provider should start with an initialization of the provider.
+
+The `initialize()` function should follow those steps:
+1. Make a GET request to the `/ofrep/v1/configuration` endpoint to retrieve the configurations return by the flag management system and store them in memory to be available for all the function of the provider. *(See [Annexe 1](#annexe-1) for the description of the endpoint response)*
```
beeme1mr, line 26: Can this step be optional? The provider should be able to work without having to call the configuration API.

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+    D --> |NO| E(return a FlagNotFound error)
+    D --> |YES| F{Is cached evaluation\nresponse in error?} 
+    F --> |YES| G(Map the error as a\nprovider error and return)
+    F --> |NO| H{Is the flag value the\nsame type as the\nevaluation function?}
+    H --> |YES| I(return the evaluation response)
+    H --> |NO| J(return a TypeMismatch error)
+```
+
+## Polling
+The polling system will make a POST request periodically tp the `/ofrep/v1/evaluate/flags` endpoint to check if there is a change in the flags evaluation to be able to store it.
```
beeme1mr, line 57:
```suggestion
The polling system will make a POST request periodically to the `/ofrep/v1/evaluate/flags` endpoint to check if there is a change in the flags evaluation to be able to store it.
```

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+### Capabilities
+In this section we will describe each capability and describe their default value and usage.
+
+#### Cache Invalidation
+The capability `cacheInvalidation` describes how the mechanism of cache invalidation works.
+
+##### Polling
+`polling`: define how the provider should do the polling
+
+- `enabled`: if `true` the provider should poll the `/ofrep/v1/evaluate/flags` regularly to check if any flag evaluation has changed in the flag management system.
```
beeme1mr, line 81: This isn't the current behavior in the OFREP web provider.

https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts#L90

guideline/static-context-provider.md
lukas-reining: This is also my general feeling, that this should be an opt-in feature as @beeme1mr says.

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
+
+## Initialize the provider
+An implementation of an OFREP client provider should start with an initialization of the provider.
+
+The `initialize()` function should follow those steps:
+1. Make a GET request to the `/ofrep/v1/configuration` endpoint to retrieve the configurations return by the flag management system and store them in memory to be available for all the function of the provider. *(See [Annexe 1](#annexe-1) for the description of the endpoint response)*
+2. Make a POST request to the `/ofrep/v1/evaluate/flags` endpoint with the evaluation context in the body.  
+   - If the endpoint returns an error, the `initialize()` function must error and exit.  
+   - If the request is successful, we should store in a local cache all of the flags evaluation results returned by the API in a local cache. We should also store the `ETag` header in the provider to be able to send it back later.
```
lukas-reining, line 29: I think it would be better to leave the `we` out and talk in third person about the provider.

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+3. If polling is enabled, the function should start the polling loop *(See [polling section](#polling))*.
```
lukas-reining, line 30:
```suggestion
3. If polling is enabled, the polling loop should start now *(See [polling section](#polling))*.
```

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Evaluation
+The evaluation should not perform any remote API calls.
+
+When calling an evaluation function the provider should check if the associated type is supported, by checking the key `capabilities.flagEvaluation.unsupportedTypes` from the configuration endpoint.
+- If the type is unsupported, we should exit early and directly return an error.
```
lukas-reining, line 36: Do we know which error we want to throw here? I think it would be nice to specify here.
Maybe we can even have a new Error type, the case that a system does not support a type happens pretty often. (independently from OFREP)

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+    A[evaluation\nfunction] --> B{Is function type in\nunsupportedTypes?}
+    B --> |YES| C(return an error)
+    B --> |NO| D{Is flag key stored\nin local cache?}
+    D --> |NO| E(return a FlagNotFound error)
+    D --> |YES| F{Is cached evaluation\nresponse in error?} 
+    F --> |YES| G(Map the error as a\nprovider error and return)
+    F --> |NO| H{Is the flag value the\nsame type as the\nevaluation function?}
+    H --> |YES| I(return the evaluation response)
+    H --> |NO| J(return a TypeMismatch error)
+```
```
lukas-reining, line 54: Love it! :)

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Polling
+The polling system will make a POST request periodically tp the `/ofrep/v1/evaluate/flags` endpoint to check if there is a change in the flags evaluation to be able to store it.
+
+If an `ETag` is available we should always add the header `If-None-Match` with the `ETag` value.
+- If the cache is still up-to-date we will receive a `304` telling us that the nothing has changed on the flag management system side.
+- If the cache is outdated we will receive a `200` with the new values of all the flags. In that situation we should:
+  1. Replace the actual local cache of flags evaluations.
+  2. Store the new `ETag` value for the future call.
```
lukas-reining, line 63:
```suggestion
If an `ETag` of a former evaluation is available we should always add the header `If-None-Match` with the `ETag` value.
- If the cache is still up-to-date we will receive a `304` telling us that the nothing has changed on the flag management system side.
- If the cache is outdated we will receive a `200` with the new values of all the flags. In that situation we should:
  1. Replace the actual local cache of flags evaluations.
  2. Store the new `ETag` value for the future call.
```

lukas-reining (commented):
Nice start and it is really valuable to have!
I left some comments, and I would generally tent to describe the expected behavior of the provider in third person instead of using first person developer.

guideline/static-context-provider.md
nicklasl: Yes, this is excellent!

guideline/static-context-provider.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Constructor
+An implementation of an OFREP client provider should allow in the creation of the provider to take as options:
+- `baseURL`: The base URL of the [flag management system](https://openfeature.dev/specification/glossary#flag-management-system). This should be the base of the URL pointing before the `/ofrep` namespace of the API.
+- `headers`: The headers to use when calling the OFREP endpoints *(ex:`Authorization`, Custom headers, etc ...)*.
+- `pollInteral`: The polling interval defining when to call again the flag management system.
+
+In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
+
+## Initialize the provider
```
nicklasl, line 22: I'm a bit torn here wether my thinking is going to deep or the use case is too fringe but bare with me:

Some applications have strict rules or goals of how fast they need to be able to show their initial screen. These apps may not have time (or do not want) to wait for network requests to finish before considering a feature flagging SDK to be initialised.

Obviously there are many ways one can solve this, but we have written mobile providers with disk persistence where we then can pass a `FetchingStrategy` to the constructor where the app owner can decide wether or not they want the `initialize()` function to block until fresh data is available or simply load any previous data from disk (and then refresh from the backend in the background). 

Firebase Remote Config explains the concepts [here](https://firebase.google.com/docs/remote-config/loading).

beeme1mr: I like the idea of various initialization strategies and the link to Firebase Remote Config is a great resource.

Perhaps one way to implement that would be to put this section under a "Fetch on load" strategy. We could start with a single strategy (and perhaps make it the default) but that would give us the ability to add additional strategies to support specific use cases.

guideline/static-context-provider.md
thomaspoignant: I was more thinking toward something that can be enabled by default and opt-out on demand.
See https://github.com/open-feature/protocol/pull/14#discussion_r1601965901

guideline/static-context-provider.md
thomaspoignant: OFREP web provider by default is doing polling.
You have a default value set if no `pollingInterval` is provided during the initialization.

https://github.com/open-feature/js-sdk-contrib/blob/c00f998227f55ca5a64b01d039ab6ec2aff18f7c/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts#L78

guideline/static-context-provider.md
thomaspoignant: Yes, this is true, should we add that to the OF spec?

guideline/static-context-provider.md
thomaspoignant: Interesting point here. I see the value of having multiple fetching strategies but I am not sure how to frame that into this document.

I see 2 options:
1. We try to define the different possible ways of doing the initialisation for the OFREP client providers, and we add them to the document.
2. We remove that block since there is some chance that it could be different to initialize something depending on the platform (web/mobile).

Do you have any preference?

.github/workflows/spec-validate.yaml
```diff
@@ -0,0 +1,20 @@
+name: OpenAPI spec validation action
```
Kavindu-Dodan, line 1: nit - a rebase should remove this diff :)

guideline/static-context-provider.md
lukas-reining: Mh yes I think we should add it to the OF spec.
I could open a PR next days, what do you think @thomaspoignant ?

thomaspoignant: It would be awesome

guideline/static-context-provider.md
nicklasl: I'm thinking that it might be most reasonable to ignore my comment about this and go ahead with a blocking network call as part of the initialize function in order to not block this document.

If we believe that the topic warrants additional discussion we can open an issue to discuss it there and amend the spec or add an appendix. wdyt?

provider/specs/client.md
beeme1mr: This will affect initial page load times, and the config API doesn't provide anything (at least not yet) that couldn't be defined explicitly by the OFREP constructor.

guideline/static-context-provider.md
thomaspoignant: As decided in the last OFREP meeting, we will get rid of the `/configuration` endpoint.
So I will resolve this conversation, because we won't have issue on how fast we can start.

thomaspoignant:
I have updated the guidelines with deleting the `/ofrep/v1/endpoint`.
I'll be interested in any review.

cc @beeme1mr @lukas-reining @jonathannorris @askpt

guideline/static-context-provider.md
```diff
@@ -0,0 +1,87 @@
...
+        "email": "foo.bar@ofrep.dev"
+      }
+    }
+   ```
+
+    - If the endpoint returns an error, the `initialize()` function must error and exit.  
+    - If the request is successful:
+      - We should store in a local cache all of the flags evaluation details returned by the API. 
+      - We should store the `ETag` header in the provider to be able to send it back later.
+2. If `pollInterval` is more than `0`, we should start a polling mechanism to periodically check for flag changes *([See polling section](#polling))*.
```
askpt, line 44: Should we define if I send an invalid `pollInterval`?

askpt (approved):
LGTM! Just a small suggestion but non-blocker.

lukas-reining (commented):
This is good @thomaspoignant! Left some comments related to wording/style.

guideline/static-context-provider.md
thomaspoignant: I am not sure what you mean by an invalid `pollInterval`?

askpt: When someone sends a `pollInterval` of negative or zero.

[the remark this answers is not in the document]
thomaspoignant: I've adapted the language style, I found it better now.

guideline/static-context-provider.md
```diff
@@ -0,0 +1,87 @@
...
+        "lastname": "bar",
+        "email": "foo.bar@ofrep.dev"
+      }
+    }
+   ```
+
+    - If the endpoint returns an error, the `initialize()` function must error and exit.  
+    - If the request is successful:
+      - Thew provider must store in a local cache all of the flags evaluation details returned by the API. 
+      - Thew provider must store the `ETag` header in the provider to be able to send it back later.
```
lukas-reining, line 43:
```suggestion
      - The provider must store in a local cache all of the flags evaluation details returned by the API. 
      - The provider must store the `ETag` header in the provider to be able to send it back later.
```

guideline/static-context-provider.md
```diff
@@ -0,0 +1,87 @@
...
+    D --> |YES| F{Is cached evaluation details in error?} 
+    F --> |YES| G(Map the error as an OpenFeature provider error and return it)
+    F --> |NO| H{Are the flag and evaluation return types the same?}
+    H --> |YES| I(return the evaluation response)
+    H --> |NO| J(return a TYPE_MISMATCH error)
+```
+
+## Polling
+The polling system will make a `POST` request periodically to the `/ofrep/v1/evaluate/flags` endpoint with the evaluation context in the body to check if there is a change in the flags evaluation to be able to store it.
+The goal is to be able to know when a flag has changed for this evaluation context.
```
lukas-reining, line 70:
```suggestion
The polling system must periodically (interval defined by `pollingInterval`) make a `POST` request to the `/ofrep/v1/evaluate/flags` endpoint with the evaluation context in the body to check if there is a change in the flags evaluation to be able to store it.
If the flags changed, the provider cache for evaluation is updated.
```

guideline/static-context-provider.md
```diff
@@ -0,0 +1,87 @@
...
+- `401`, `403`: The provider is not authorized to call the OFREP API. In that situation the provider should emit a `fatal` event and stop the polling mechanism. The provider will be in a [terminal error state](https://openfeature.dev/specification/sections/flag-evaluation#requirement-177).
+- `429`: The provider has reached the rate limit of the flag management system. In that situation the provider should read the `Retry-After` header from the response and ensure that no call to the endpoint happened before this date.
+- `200`: The cache is outdated. In that situation the provider must:
+  1. Clear the actual cache.
+  2. Cache the new evaluation details received.
+  3. Read the `ETag` header and store it for future call the API.
+  4. Emit an `ConfigurationChanged` event containing in the `flagsChanged` field, a list of the flag key that have changed.
+
+## Change context
+In the client providers, a change of context should be handle to retrieve a new version of the cache, it is called that the [Provider context reconciliation](https://openfeature.dev/specification/sections/providers/#26-provider-context-reconciliation).
```
lukas-reining, line 85: I do not fully get this sentence.

lukas-reining (approved):
Left 3 more comments, again only language.
Thanks Thomas!

beeme1mr:
https://github.com/open-feature/protocol/pull/42#pullrequestreview-2829849085

This feedback also applies here. The timeout is arguably more important in this context.

guideline/static-context-provider.md
```diff
@@ -0,0 +1,88 @@
...
+## Configuration
+An OFREP client provider implementation must, at the time of creation, accept at least these options
+- `baseURL`: The base URL of the [flag management system](https://openfeature.dev/specification/glossary#flag-management-system).  
+  This must be the base of the URL pointing before the `/ofrep` namespace of the API.
+  - In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
+- `headers`: The headers to use when calling the OFREP endpoints *(e.g.:`Authorization`, Custom headers, etc ...)*.
+- `pollInterval`: The polling interval defining how often to update the cached flags in the provider.
+  - If `pollInterval` is equal to 0, polling will be disabled.
+  - See [polling section](#polling) for more details.
+- `timeout`: This specifies the duration to wait for a response before canceling the HTTP request. If no value is provided, a default timeout of `10 seconds` must be applied.
```
beeme1mr, line 21:
```suggestion
- `timeoutMs`: This specifies the duration to wait for a response before canceling the HTTP request. If no value is provided, a default timeout of `10_000 milliseconds` must be applied.
```

It's called `timeoutMs` already in the JavaScript providers.

beeme1mr (approved):

guideline/static-context-provider.md
beeme1mr: No longer applicable

PR #19, opened 2024-05-28
feat: reversed lack of support to support capabilities

grimly:
## This PR

1. Reverses the `unsupportedTypes` to `supportedTypes`
2. Removes enumeration of supported types
3. Restores the required `name` in the `configurationResponse`

### Related Issues

N/A

### Notes

1. It is not a capability to not support something.
   Instead the response should express its ability to support such types.
   I reversed the naming of this capability because as this specification updates, this would force providers either to immediately show a new type to be unsupported or to immediately support a new type.
2. I removed the limitations of supported types. Adding values to an enum in a response is a breaking change. The goal is to allow a new type in the OpenFeature specification while limiting breaking changes on this OpenAPI specification.
3. I think this `required` clause was wrongly moved. I'm merely bringing it back up.
   I would nevertheless question the requirement of this property as it does not add any significant value to a client.

### Follow-up Tasks

N/A

### How to test

N/A

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -335,18 +335,21 @@ components:
               $ref: '#/components/schemas/featureCacheInvalidation'
             flagEvaluation:
               $ref: '#/components/schemas/flagEvaluation'
+      required:
+        - name
     flagEvaluation:
       type: object
       description: Configurations specific for flag evaluations in OFREP provider implementation
       properties:
-        unsupportedTypes:
-          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value.
+        supportedTypes:
+          description: |
+            A list of supported types by the flag management system.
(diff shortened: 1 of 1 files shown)


thomaspoignant:
We had a discussion with the community about this and went on purpose for `unsupportedTypes` to be sure that the default value is to accept everything.

We are missing an ADR about this, I will open an issue to add one about this. 
We will probably not reverse this logic.

Pinging @Kavindu-Dodan @lukas-reining @beeme1mr to have confirmation.

thomaspoignant:
Issue to add the ADR is available here: https://github.com/open-feature/protocol/issues/21

Kavindu-Dodan:
This overlaps with https://github.com/open-feature/protocol/pull/24

Kavindu-Dodan:
@grimly #24  reverted the unsupported types to supported types. I will close this PR as a cleanup.

PR #20, opened 2024-05-28
feat: Allow any reason

grimly:
## This PR

- Removes restriction on a successful evaluation `reason`

### Notes

The OF specification defines a [Resolution Details](https://github.com/open-feature/spec/blob/main/specification/types.md#resolution-details) with a "reason [that] should not be limited to the reasons enumerated above. It can be any of the pre-defined reasons, or any string value."

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -210,7 +210,7 @@ components:
               $ref: "#/components/schemas/key"
             reason:
               type: string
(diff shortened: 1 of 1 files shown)


thomaspoignant (approved):
Good catch 👍

Kavindu-Dodan (approved):

lukas-reining (approved):

beeme1mr (approved):

PR #22, opened 2024-05-30
feat: Remove name from configuration response

grimly:
## This PR

- Removes name from the `GET /ofrep/v1/configuration` response

### Notes

This property provides no added value to the API client.

While I understand why a vendor want to promote his product by name, this is the role of the [`Server` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server).

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -321,12 +321,6 @@ components:
     configurationResponse:
       description: OFREP metadata response
       properties:
-        name:
-          type: string
-          description: name of the flag management system
-          examples:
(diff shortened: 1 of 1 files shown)


grimly:
You can put this information on the `Server` header.
It can even be sent on all evaluations, therefore providing this information even on those traces

Kavindu-Dodan:
I think `name` can be an optional property. While I agree this could be resolved with headers, that adds some extra complexity at the OFREP implemetaiton.

PR #24, opened 2024-06-07
feat!: Reverse the logic for supportedTypes

thomaspoignant:
## This PR
This PR reverse in the `configuration` endpoint how we specify the supported types.

⚠️ Breaking change: since we are still in WIP mode for OFREP we accept this breaking change.
We change the name of the field from `unsupportedTypes` to `supportedTypes`.
If not set the default value is to support all the types.

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -13,46 +13,24 @@ info:
 security:
   - ApiKeyAuth: [ ]
   - BearerAuth: [ ]
+tags:
+  - name: OFREP Core
+    description: |
+      **Required**: Core APIs to implement to support OFREP.  
+      *This is the minimum set of APIs required for a flag management system to be OFREP compatible.*
+  - name: OFREP Extensions
+    description: |
+      **Optional**: Extension APIs to provide full support for OFREP.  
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -339,14 +339,17 @@ components:
       type: object
       description: Configurations specific for flag evaluations in OFREP provider implementation
       properties:
-        unsupportedTypes:
-          description: A list of unsupported types by the flag management system. Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value.
+        supportedTypes:
+          description: |
+            A list of supported types by the flag management system.  
+            Evaluating a flag of a listed type through OFREP provider will result in an error and yield the default value. Undefined means that the flag management system is supporting all types.
```
Kavindu-Dodan, line 345:
```suggestion
            Evaluating a flag of unlisted type through the OFREP provider will result in an error and yield the default value. However, when supportedTypes is undefined/empty, provider assumes that all flag evaluation types are supported by the flag management system
```

Kavindu-Dodan (approved):
+1 for switching to supported types

lukas-reining (approved):

PR #25, opened 2024-06-07
feat!: change minPollingInterval field name to mention millisecond

thomaspoignant:
## This PR
This PR renames in the `configuration` endpoint the `minPollingInterval` to `minPollingIntervalInMs` to be explicit to API users that the value expected should be in milliseconds.

⚠️ Breaking change: since we are still in WIP mode for OFREP we accept this breaking change.

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -360,9 +360,11 @@ components:
         enabled:
           type: boolean
           description: set to true if the remote flag management system is supporting polling
-        minPollingInterval:
+        minPollingIntervalInMs:
           type: number
-          description: minimum polling interval (in millisecond) supported by the flag management system. The provider should ensure not to set any polling value under this minimum.
+          description: |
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -360,9 +360,11 @@ components:
         enabled:
           type: boolean
           description: set to true if the remote flag management system is supporting polling
-        minPollingInterval:
+        minPollingIntervalInMs:
```
beeme1mr, line 363:
```suggestion
        minPollingIntervalMs:
```

This is another option but I could go either way as long as we stay consistent.

beeme1mr (approved):

Kavindu-Dodan (approved):

service/openapi.yaml
Kavindu-Dodan: I think `minPollingIntervalMs` is better

PR #34, opened 2025-01-22
feat: add flag set metadata for bulk response and failures

toddbaert:
This PR adds the existing metadata flag prop to the bulk evaluation and to failures.

These cases are particularly important for telemetry (see our recent addition of appendix D). Even in error cases, flag metadata (such as the version, flag set id, etc) can be very useful in debugging and metrics. These additions facilitate these use cases, and are non-breaking and backwards compatible.

The existing `metadata` struct has been moved to the components section for reuse.

image

image

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -201,6 +201,8 @@ components:
       required:
         - flags
       properties:
+        metadata:
+          $ref: "#/components/schemas/metadata"
         flags:
           type: array
           items:
@@ -244,13 +246,7 @@ components:
               type: string
               description: Variant of the evaluated flag value
             metadata:
-              type: object
-              additionalProperties:
-                oneOf:
-                  - type: boolean
-                  - type: string
-                  - type: number
-              description: Arbitrary metadata supporting flag evaluation
+              $ref: "#/components/schemas/metadata"
         - oneOf:
             - $ref: "#/components/schemas/booleanFlag"
             - $ref: "#/components/schemas/stringFlag"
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -201,6 +201,8 @@ components:
       required:
         - flags
       properties:
+        metadata:
+          $ref: "#/components/schemas/metadata"
```
beeme1mr, line 205: Please move this after the flags property.

toddbaert: Done

service/openapi.yaml
```diff
@@ -421,3 +434,17 @@ components:
...
+        oneOf:
+          - type: boolean
+          - type: string
+          - type: number
+    flagMetadataExamples:
+      examples:
+        - team: ecommerce
+          businessPurpose: experiment 
+    flagMetadataDescription:
+      description: Arbitrary metadata for the flag, useful for telemetry and documentary purposes.
```
beeme1mr, line 449: Couldn't you simplify this by doing something like:

```yaml
flagMetadata:
  $ref: "#/components/schemas/metadata"
  description: Arbitrary metadata for the flag, useful for telemetry and documentary purposes.
  examples:
        - team: ecommerce
          businessPurpose: experiment 
```

Then you could get rid of the `allOf` and just reference `#flagMetadata`

beeme1mr (approved):

lukas-reining (approved):

service/openapi.yaml
beeme1mr: Never mind, this doesn't seem to be valid in OpenAPI.

PR #36, opened 2025-02-07
ci: switch OpenAPI validators

beeme1mr:
I looked into using https://github.com/daveshanley/vacuum but it doesn't have a GitHub action yet.

### DIFF
--- a/.github/workflows/spec-validate.yaml
+++ b/.github/workflows/spec-validate.yaml
@@ -15,6 +15,6 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run OpenAPI spec validation
-        run: |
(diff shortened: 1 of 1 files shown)


lukas-reining (approved):

.github/workflows/spec-validate.yaml
```diff
@@ -1,7 +1,7 @@
 name: OpenAPI spec validation action
 
 on:
-  pull_request_target:
+  pull_request:
```
beeme1mr, line 4: I believe this should be `pull_request` instead of `pull_request_target`.

lukas-reining: Yes you are right, it has to run on the merge commit and not on the base branch, so  pull_request is correct.

PR #42, opened 2025-05-08
feat: Server provider guidelines

### LINKED ISSUE #41: Write server provider guidelines
Write the guidelines on how to create a server OFREP provider.

This document should help contributors to create an OFREP provider for any server providers.

thomaspoignant:
## This PR
This PR contains the guidelines on how to create a server provider for OFREP.

Closes #41

### DIFF
--- a/guideline/dynamic-context-provider.md
+++ b/guideline/dynamic-context-provider.md
@@ -1,0 +1,39 @@
+# Creating an OFREP server provider
+
+OpenFeature Remote Evaluation Protocol (OFREP) is an API specification for feature flagging that allows the use of generic providers to connect to any feature flag management systems that supports the protocol.
+
+In this document, we will specify how to write an OFREP provider using the [dynamic-context-paradigm](https://openfeature.dev/specification/glossary/#dynamic-context-paradigm) that is used on server side applications typically operate in the context of multiple users. 
+
+**Pre-requisite:**
+- Understanding of [general provider concepts](https://openfeature.dev/docs/reference/concepts/provider/)
+- Understanding of the [OFREP](../../README.md)
+- Understanding of the [OFREP OpenAPI specification](../../service/openapi.yaml)
+
+## Configuration
+An OFREP server provider implementation must, at the time of creation, accept at least these options
+- `baseURL`: The base URL of the [flag management system](https://openfeature.dev/specification/glossary#flag-management-system).  
+  This must be the base of the URL pointing before the `/ofrep` namespace of the API.
+  - In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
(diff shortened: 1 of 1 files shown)


guideline/dynamic-context-provider.md
```diff
@@ -0,0 +1,40 @@
...
+**Pre-requisite:**
+- Understanding of [general provider concepts](https://openfeature.dev/docs/reference/concepts/provider/)
+- Understanding of the [OFREP](../../README.md)
+- Understanding of the [OFREP OpenAPI specification](../../service/openapi.yaml)
+
+## Constructor
+An implementation of an OFREP server provider should allow in the creation of the provider to take as options:
+- `baseURL`: The base URL of the [flag management system](https://openfeature.dev/specification/glossary#flag-management-system).  
+  This should be the base of the URL pointing before the `/ofrep` namespace of the API.
+  - In the constructor, the provider should check if the `baseURL` is a valid URL and return an error if the URL is invalid.
```
askpt, line 17: Is this necessary? The server doesn't need to know its own URL. I am confused here.

askpt (commented):
Small comment, otherwise looks good.

guideline/dynamic-context-provider.md
lukas-reining: This is the OFREP server and not the "client server" doing feature flagging.

lukas-reining (commented):
This is great, the feedback from the client provider applies here too @thomaspoignant :)

guideline/dynamic-context-provider.md
thomaspoignant: Exactly what Lukas said, the provider has to know which flag management system to use.

thomaspoignant:
@lukas-reining I have adapted the same changes to the server provider guideline.

askpt (approved):

lukas-reining (approved):

beeme1mr (approved):
We may want to briefly mention http timeout behavior. In the web provider I believe we default to 10 seconds and the provider is put in an error state.

It may also be worth briefly describing the expected behavior if the response is invalid. I would expect a provider error if it occurs during initialization. If the provider receives an unexpected response after initialization I would expect the provider to use the previously cached value and possibly emit an error event.

thomaspoignant:
> We may want to briefly mention http timeout behavior. In the web provider I believe we default to 10 seconds and the provider is put in an error state.


Yes mentioning the timeout is a good idea I will add it.


> It may also be worth briefly describing the expected behavior if the response is invalid. I would expect a provider error if it occurs during initialization. If the provider receives an unexpected response after initialization I would expect the provider to use the previously cached value and possibly emit an error event. 

There is no caching in the server provider so I am not sure what you mean here?
This is something we can add but I am not sure if it is in scope for the `1.0`.

PR #43, opened 2025-05-16
feat!: remove cacheable property

thomaspoignant:
## This PR
Remove `cacheable` property because there is no plan to support cache in the providers before `1.0`.


beeme1mr (approved):

lukas-reining:
Just to be sure: the reason for removing is because for now we are not going to cache in the server? @thomaspoignant @beeme1mr

PR #49, opened 2025-07-29
feat!: add code default flag support

beeme1mr:
Update the OpenAPI schema to support a code default flag. This is the first step in implementing the plan defined in [this ADR](https://github.com/open-feature/protocol/blob/main/service/adrs/0006-optional-value-for-code-defaults.md).

### DIFF
--- a/service/openapi.yaml
+++ b/service/openapi.yaml
@@ -2,7 +2,7 @@ openapi: 3.1.0
 servers:
   - url: /
 info:
-  version: 0.1.0
+  version: 0.2.0
   title: OpenFeature Remote Evaluation Protocol (OFREP)
   description: OFREP define the protocol for remote flag evaluations
   contact:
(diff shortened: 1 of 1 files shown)


service/openapi.yaml
```diff
@@ -2,7 +2,7 @@ openapi: 3.1.0
 servers:
   - url: /
 info:
-  version: 0.1.0
+  version: 0.2.0
```
beeme1mr, line 5: I'm not sure if we want to bump this version or not.

lukas-reining: We did not do it until now, but it makes sense to me as this is "breaking".

beeme1mr: Cool, thanks. Do you have any concerns with the PR?

askpt (approved):
LGTM!

lukas-reining (approved):
Looks good to me!

service/openapi.yaml
beeme1mr: https://github.com/open-feature/protocol/issues/50

toddbaert (approved):

Issue #62, opened 2026-02-20
OFREP SSE Support


jonathannorris:
## Problem

OFREP currently only supports polling for flag change detection in client-side (static context) providers. There is no way to implement real-time flag updates, and frequent polling introduces unnecessary load on flag management systems while adding latency to flag change propagation. Many vendors already use SSE for change notification, but without a standardized mechanism in OFREP, each must implement proprietary push solutions.

## Proposed Solution

Add an optional `sse` array to the bulk evaluation response (`POST /ofrep/v1/evaluate/flags`). When present, it provides SSE endpoint URLs that the provider connects to for real-time flag change notifications. SSE events contain a `type` field (e.g., `refetchEvaluation`) along with optional `etag` and `lastModified` values that the provider uses when re-fetching the bulk evaluation via the existing ETag-based caching mechanism.

This is a backward-compatible, additive change -- servers that don't support SSE omit the field, and providers that don't support it continue polling.

Further details on schema, event format, provider behavior guidelines, and open questions will be provided in the [ADR PR](https://github.com/open-feature/protocol/pull/63).

PR #63, opened 2026-02-20
docs: add ADR for SSE bulk evaluation change notifications

jonathannorris:
## This PR

This PR adds ADR-0008 (#62) to propose Server-Sent Events (SSE) for bulk evaluation change notifications in OFREP.

- Adds an optional `refreshConnections` array to bulk evaluation responses, with a `type` discriminator to support future push mechanisms beyond SSE.
- Defines notification-only SSE behavior (`refetchEvaluation`) and provider lifecycle guidance (reconnect, fallback, coalescing).
- Specifies SSE-trigger metadata transport via query params (`sseEtag`, `sseLastModified`) **only** for requests triggered directly by SSE messages.

### Notes

Includes open questions on query params vs custom headers, replay guarantees (`Last-Event-ID`), and security expectations for tokenized SSE URLs.


service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,244 @@
...
+- There is no way to implement real-time flag updates
+- Frequent polling introduces unnecessary load on flag management systems
+- There is an inherent latency between flag changes and client awareness, bounded by the poll interval
+
+The [vendor survey](https://docs.google.com/forms/d/1NzqKx57XvRK_2lRQOFCRmF5exet6f15-sCjdEy0HCS8#responses) referenced in ADR-0005 confirmed that many vendors already use SSE for change notification. Without a standardized mechanism in OFREP, each vendor must implement proprietary push solutions, undermining the protocol's goal of vendor-agnostic interoperability.
+
+Server-Sent Events (SSE) is a W3C standard that fits this use case well:
+- Unidirectional (server-to-client), matching the notification pattern
+- Runs over standard HTTP without protocol upgrades
+- Natively supported in browsers via the `EventSource` API
```
askpt, line 25: I would consider adding that mobile is also supported. After a quick search I confirmed that.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,244 @@
...
+{
+  "flags": [
+    {
+      "key": "discount-banner",
+      "value": true,
+      "reason": "TARGETING_MATCH",
+      "variant": "enabled"
+    }
+  ],
+  "sse": [
```
askpt: Not a big fan of this object. I prefer to not use the name of the technology here, but its purpose.
For example:

```json
"refresh": [ //I am terrible with naming so I am up for suggestions
{
"type": "sse",
"url": "http://",
"timeout": 123
}]
```

This would allow us to add extra notification/refresh type without breaking the contract.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,244 @@
...
+- **Connection resource usage**: Long-lived SSE connections consume resources on both client and server, particularly at scale
+- **Re-fetch amplification risk**: Multiple SSE URLs or bursty event streams can trigger redundant concurrent re-fetches unless providers coalesce events
+- **Transport consistency trade-off**: Using query parameters for SSE metadata differs from common HTTP conditional request patterns and may need careful documentation for implementers
+- **Tokenized URL handling risk**: If SSE URLs include scoped credentials or channel tokens, accidental logging/persistence can expose sensitive connection material
+
+## Open Questions
+
+1. **Should `refetchEvaluation` be required, or should providers refetch on any SSE message?** Requiring a specific `type` field enables future event types without triggering unnecessary refetches. Refetching on any message is simpler. This ADR recommends requiring `type=refetchEvaluation` for forward compatibility.
+2. **Should providers support streaming full evaluation payloads over SSE?** This ADR focuses on the notification pattern. Full payload streaming could be specified as a separate event type in a future revision.
```
askpt: I believe as a potential v2 for SSE, I would like to experiment having a json PATCH similar behaviour. What I mean is, instead of forcing a full refresh, the event would send the flags that need to be updated.

For example:
```json
[
{ "op": "add", flagKey: "test", value: { } },
{ "op": "replace", flagKey: "test", "path": "/defaultValue", "value": "false" },
{ "op": "remove", flagKey: "test"}
]

```

This adds extra complexity but it would be a good improvement.

askpt (commented):
Added a couple of comments for discussion.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,244 @@
...
+### Negative
+
+- **Additional provider complexity**: Providers must manage SSE connection lifecycle, reconnection, inactivity handling, and fallback behavior based on configured change detection settings
+- **Infrastructure requirements**: Flag management systems that want to support SSE need to operate or integrate with an SSE-capable service
+- **Connection resource usage**: Long-lived SSE connections consume resources on both client and server, particularly at scale
+- **Re-fetch amplification risk**: Multiple SSE URLs or bursty event streams can trigger redundant concurrent re-fetches unless providers coalesce events
+- **Transport consistency trade-off**: Using query parameters for SSE metadata differs from common HTTP conditional request patterns and may need careful documentation for implementers
+- **Tokenized URL handling risk**: If SSE URLs include scoped credentials or channel tokens, accidental logging/persistence can expose sensitive connection material
+
+## Open Questions
```
beeme1mr, line 295: Should the developer be able to opt out of SSE via a config option?

jonathannorris: well SSE is totally optional, it basically comes down to if the Server responds with an `sse` object or not. But yes the Providers should also have a disable SSE option in them.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: yea Mobile should be supported with SSE will make that clearer. I recommend that we stick to the LaunchDarkly Event Source libraries where its not built-in: 
- https://www.npmjs.com/package/launchdarkly-eventsource
- https://github.com/launchdarkly/dotnet-eventsource
- https://github.com/launchdarkly/swift-eventsource
- https://github.com/launchdarkly/okhttp-eventsource

service/adrs/0008-sse-for-bulk-evaluation-changes.md
askpt: I suggest allowing the developer to choose as a setting:
- none
- pooling
- sse

This would give more control to the developer to pick up their preferred way for refreshing the flags.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: yea I agree, going to use `refreshConnection` with a type of `sse` instead of `sse` directly.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: yea makes sense, but yea I would leave that for a new ADR.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,244 @@
...
+3. **What is the recommended coalescing strategy when multiple SSE connections are specified?** Providers should connect to all URLs, but should re-fetch in a coalesced way (debounce + in-flight dedupe) to avoid amplification. Should OFREP define minimum coalescing expectations?
+4. **Should `inactivityDelaySec` be server-provided or client-side configuration?** This ADR specifies it as server-provided, allowing the server to tune connection lifecycle. Providers may also expose a client-side override.
+5. **Should non-`refetchEvaluation` SSE messages be forwarded to the provider?** Should we add a mechanism to support non-`refetchEvaluation` typed messages that are forwarded through to the provider via an events/hook interface?
+6. **Should SSE metadata be transported via query parameters or custom headers?** This ADR currently recommends query params (`sseEtag`, `sseLastModified`) due to browser CORS preflight considerations and the non-conditional-request semantics. Should OFREP also define an optional custom header form for non-browser clients?
```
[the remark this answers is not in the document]
jonathannorris: We considered allowing custom headers as an alternative for non-browser clients, but opted to keep query params as the single transport mechanism for consistency — having two code paths (headers for server/mobile, query params for browser) adds implementation complexity without a meaningful benefit. The CORS preflight concern is the primary driver for browser clients, but query params work equally well for mobile and server-side SSE consumers. Keeping a single mechanism makes the spec simpler to implement correctly across all SDK types.

erka:
It would be helpful to include an error scenario as well. At the moment, the ADR focuses only on the happy-path flow.

For example, consider a data center issue where some pods have already been updated while others have not. The client receives a refresh event over SSE, but the subsequent refetch request is routed to a pod that has not yet been updated. How should this situation be handled?

beeme1mr (approved):

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+OFREP currently relies exclusively on polling for flag change detection in client-side (static context) providers. As described in [ADR-0005](0005-polling-for-bulk-evaluation-changes.md), polling was chosen initially for simplicity, with the explicit expectation that additional change detection mechanisms would be added later.
+
+This ADR focuses primarily on static-context providers (for example web and mobile SDK providers) that use bulk evaluation caching patterns. It does not introduce SSE requirements for dynamic-context providers that primarily use single-flag evaluations.
```
lukas-reining: Why only for `static-context`?
While I agree that this is the much more _needed_ use case, the [PROVIDER_CONFIGURATION_CHANGED](https://openfeature.dev/docs/reference/concepts/events/#provider_configuration_changed) event would still be good to support for OFREP in dynamic context scenarios with this addition.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: @jonathannorris can you please add that there are implementations for all mobile platforms supported by OF as suggested by @askpt?

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,270 @@
...
+Server-Sent Events (SSE) is a W3C standard that fits this use case well:
+- Unidirectional (server-to-client), matching the notification pattern
+- Runs over standard HTTP without protocol upgrades
+- Natively supported in browsers via the `EventSource` API
+- Built-in reconnection support; when events include `id`, reconnecting clients can send `Last-Event-ID` to resume from the last processed event
+- Works through proxies, CDNs, and standard HTTP infrastructure
+
+## Decision
+
+Add an optional `refreshConnections` array to the bulk evaluation response (`POST /ofrep/v1/evaluate/flags`). When present, it provides connection endpoints that the provider connects to for real-time flag change notifications.
```
lukas-reining: I do not find `refreshConnections` very speaking, but this might just be an opinion of mine. 
To me something like `providerStaleneChannels` or `providerConfiguarationChangedChannels`, would be more intuitive and reflect the spec. Even though I am not happy with `channel` but I could not come up with something better yet. 

https://openfeature.dev/docs/reference/concepts/events/#provider_stale

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,270 @@
...
+This is primarily intended for static-context providers (for example web/mobile clients) that rely on bulk evaluations.
+
+SSE is used as a **notification-only** mechanism -- events signal the provider to re-fetch the bulk evaluation via the existing endpoint, rather than streaming full evaluation payloads. This keeps the SSE message format simple, reuses existing infrastructure, and avoids duplicating evaluation logic.
```
lukas-reining: I am going back and forth with this. Your arguments for this make sense. 

But for many clients this would still mean huge amount of requests coming in parallel if all providers are notified in parallel. You could delay the events but that would contradict our effort to enable real time update. This can also become complex. 

Sending the diff (or all flags if wanted) would make this much more predictable in my view. The downside would be so decide which flag to update but maybe simple `UPSERT` or per flag instructions would make it.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,270 @@
...
+  ],
+  "metadata": {
+    "version": "v12"
+  }
+}
+```
+
+Each refresh connection object has:
+- `type` (string, required): The connection type. Currently `"sse"` is the only defined value. Providers must ignore entries with unknown types for forward compatibility, allowing new push mechanisms to be added without breaking existing clients.
+- `url` (string, required): The endpoint URL. The URL is opaque to the provider and may include authentication tokens, channel identifiers, or other vendor-specific query parameters. Implementations must treat this URL as sensitive -- it may contain auth tokens or channel credentials -- and must not log or persist the full URL including query string.
```
lukas-reining: The opaqueness is a very good idea!

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: Tbh., to me this would be the way out of the box as described above.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: I'm curious what the use-case looks like here for dynamic-context providers, since each flag evaluation for dynamic context would be a fresh OFREP request with its own evaluation context, there's no local cache to invalidate or refetch. What would the `PROVIDER_CONFIGURATION_CHANGED` signal to a server-side provider that isn't already handled by the next evaluation request returning fresh values?

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: yea I'm also struggling with the naming on this, originally I just had it `sse`, but trying to find a more generic name that fits. What are your thoughts on using one of these: 
- `subscriptions`
- `liveConnections`
- `channels`

cc @toddbaert if you have any ideas

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,270 @@
...
+
+    Note over SSE,Client: Real-time change notification
+    SSE-->>Client: event: refetchEvaluation (etag, lastModified)
+    Client->>Server: POST /ofrep/v1/evaluate/flags?sseEtag=etag&sseLastModified=lastModified
+    alt Flags changed
+        Server-->>Client: 200 OK (new flags + ETag)
+        Client->>Client: Update cache, emit ConfigurationChanged
+    else Flags unchanged
+        Server-->>Client: 304 Not Modified
+    end
```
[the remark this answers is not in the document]
jonathannorris: This is intentional. SSE-triggered re-fetches omit `If-None-Match` — the `sseEtag` query parameter serves a different purpose: it carries the ETag from the SSE event payload so the server can validate its internal cache state (e.g. against object storage), not to validate the client's cached response. The two mechanisms are kept separate to avoid conflating client-side cache validation with server-side freshness checks triggered by the SSE channel.

jonathannorris:
> It would be helpful to include an error scenario as well. At the moment, the ADR focuses only on the happy-path flow.
> 
> For example, consider a data center issue where some pods have already been updated while others have not. The client receives a refresh event over SSE, but the subsequent refetch request is routed to a pod that has not yet been updated. How should this situation be handled?

Yea @erka this is a real edge case we've encountered in production at DevCycle running the same SSE + CDN cache pattern at scale.

In practice we handle this entirely within the worker layer and very rarely see issues. When the client re-fetches after receiving an SSE notification, it passes the `sseEtag` and `sseLastModified` values from the SSE event as query parameters. The worker uses these to validate the response it gets back from the CDN — if the response is older than what the SSE event said the new config should be, the worker invalidates the cache and retries until it fetches the updated value.

It's worth noting this is specific to how we're looking to implement it based on our experience with this pattern — the `sseEtag` and `sseLastModified` query params are intentionally left flexible in the ADR so different platforms and infrastructure setups can use them in whatever way makes sense for their consistency guarantees.
We're planning to implement this behaviour in https://github.com/open-feature/flagd-ofrep-cf-worker as a reference implementation.

I've added an error/stale scenario to the sequence diagram to document this flow more explicitly.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: The notification-only + re-fetch approach is closely tied to how we've implemented this at DevCycle, where the architecture relies on CDNs and edge workers designed to handle the burst of concurrent re-fetch requests that fire when an SSE event reaches all connected clients simultaneously. In that setup the CDN absorbs the load naturally and the edge worker handles cache consistency internally.

For providers without a CDN or edge layer, sending the full flag payload or a JSON patch over SSE may be a better fit and avoids the thundering herd problem entirely. The `type` field in the event payload is designed to be extensible for exactly this reason, so a future ADR version could define a `patchConfig` or `fullConfig` event type without breaking the existing contract.

We're planning to build out the refetch pattern as a reference implementation in https://github.com/open-feature/flagd-ofrep-cf-worker.

askpt (approved):

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: Okay this makes sense to me. The question here is, which is the most common case. 
What you describe relies on much abstracted cloud services, which in my view does not even count for a "standard" K8s based architecture, where such elasticity to survive a thundering herd case is not given. 

Anecdotally I can say that, this is the kind of setup most of my clients of the past have been using. 
And with thousands of online customers, this can be a problem elasticity wise. 

> The type field in the event payload is designed to be extensible for exactly this reason, so a future ADR version could define a patchConfig or fullConfig event type without breaking the existing contract.

This is very good!

>  so a future ADR version could define a patchConfig or fullConfig event type without breaking the existing contract

I would like to discuss to have this included here to have both specified from the beginning. 
What do you think? If not, I am open to open a parallel ADR but I think it would be better to have both worlds solved in one go.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: In servers, not all FFs are evaluated during a request. Some are simply for the server. 
For these, you want to be notified about the flag value changing, independent from a running request and caller context. 

There are cases with longer lasting connections e.g. for SSEs or WebSockets where a server SDK with dynamic-context is used but FFs are evaluated.
In these cases we have been using events in OF to be notified about this. 

As said, I agree that the static-context case is much more common.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: Yeah I was also thinking pretty hard on this :D 
I feel like `subscriptions` could be okay and kind of natural, as long as we agree that flag updates are the only thing we can naturally subscribe to. 
To me including `stalenessSubscriptions` or `flagSetSubscriptions` could also make sense to be future proof and make this more specific to what we use this for.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
toddbaert: I think I agree with @lukas-reining - we do have backend OFREP implementations, and `PROVIDER_CONFIGURATION_CHANGED` are useful (though admittedly less common) for server-side scenarios. We have services at Dynatrace using event driven models to control things like log levels, and other "non-user" operational flags that change the application state. Getting async messages about these changes is something flagd supports, even in server scenarios, and I suppose the same could be done in an OFREP implementation.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,291 @@
...
+- There is an inherent latency between flag changes and client awareness, bounded by the poll interval
+
+The [vendor survey](https://docs.google.com/forms/d/1NzqKx57XvRK_2lRQOFCRmF5exet6f15-sCjdEy0HCS8#responses) referenced in ADR-0005 confirmed that many vendors already use SSE for change notification. Without a standardized mechanism in OFREP, each vendor must implement proprietary push solutions, undermining the protocol's goal of vendor-agnostic interoperability.
+
+Server-Sent Events (SSE) is a W3C standard that fits this use case well:
+- Unidirectional (server-to-client), matching the notification pattern
+- Runs over standard HTTP without protocol upgrades
+- Natively supported in browsers via the `EventSource` API
+- Built-in reconnection support; when events include `id`, reconnecting clients can send `Last-Event-ID` to resume from the last processed event
+- Works through proxies, CDNs, and standard HTTP infrastructure
```
toddbaert, line 27:
```suggestion
- Works through proxies, CDNs, and standard HTTP infrastructure
- Can be represented in OpenAPI spec with `text/event-stream`
```

I think in the latest versions, the representation is good enough that code generation even works (I think we need 3.2.0 for this though, and I believe we are at [3.1.0](https://github.com/open-feature/protocol/blob/main/service/openapi.yaml#L1))

service/adrs/0008-sse-for-bulk-evaluation-changes.md
toddbaert: If we are keeping this generic, what doing `eventStreams` or something like that? Or are the semantics of the request/response here too narrow for this general use? If not, we can then capture all of the specific semantics in the types and metadata of the events themselves and make this extensible.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
toddbaert: @lukas-reining if you have some ideas for a extensions to support other approaches, I think adding them in this PR is fine if they can be captured concisely. Otherwise, maybe you can just see if the extensibility @jonathannorris has proposed is sufficient and do a follow up (I'm happy to help with this).

I see your concerns about "thundering herd" and it totally makes sense in some cases, but due to the decentralized architecture we are working on it's less of a realistic problem for us... but also for that reason I don't think we're the best people to enhance this proposal to fit some of the architectures you are familiar with. It also seems to me that things like jitter, etc could help here, and that's probably more of an implementation than an API question.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,291 @@
...
+SSE envelope fields:
+- `id` (string, recommended): Event identifier used by SSE clients for resume semantics via `Last-Event-ID`.
+
+Reconnection and replay behavior:
+- Providers should rely on standard SSE reconnect behavior and pass `Last-Event-ID` when supported by the client/runtime.
+- Servers that support replay should emit stable event `id` values for `refetchEvaluation` events and replay missed events when `Last-Event-ID` is provided.
+- Providers must perform an immediate bulk re-fetch after reconnect, even when replay is supported, to guarantee cache correctness across implementations with different replay retention policies.
+
+Transporting SSE metadata to the bulk endpoint:
+- `sseEtag` and `sseLastModified` are SSE-trigger metadata, not standard HTTP conditional request validators for endpoint-level response caching semantics.
```
toddbaert: @jonathannorris IMO, `sseEtag` and `sseLastModified` sound like they are etags FOR the SSE connection, not FROM it. I think something like `flagConfigEtag` or something in that direction would be clearer, and leave less of a semantic  gap between the associated HTTP `If-None-Match` headers, which are about the payload/config entity.

I should add this is a bit of a naming nit, so if you think I'm way off about this, please disregard.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,291 @@
...
+
+4. **Should `inactivityDelaySec` be server-provided or client-side configuration?**
+   - **Answer:** This ADR specifies `inactivityDelaySec` as server-provided, defaulting to 120 seconds when omitted. Providers may expose a client-side override, which should take precedence over the server-provided value.
+
+5. **Should non-`refetchEvaluation` SSE messages be forwarded to the provider?**
+   - **Answer:** A mechanism to forward unknown typed messages to the provider via an events/hook interface could be valuable but is deferred to a future revision.
+
+6. **Should SSE metadata be transported via query parameters or custom headers?**
+   - **Answer:** Query params are used as the single transport mechanism for all SDK types. Custom headers were considered but rejected because non-safelisted headers require expanding `Access-Control-Allow-Headers` in CORS configuration, and introduce additional complexity for browser-based SDKs. Query params also make the SSE origin of the metadata explicit, distinguishing `sseEtag`/`sseLastModified` from standard HTTP conditional request headers (`If-None-Match` / `If-Modified-Since`).
```
toddbaert: Another point in favor of this reasoning is that, at least thus far, we've put "application-level" data in path, query, body params and metadata and "implementation concerns" like caching, etc, in header fields. I think SSE falls more into the first category since based on this proposal it can deliver arbitrary event data related to the domain of feature flagging.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,291 @@
...
+      "key": "discount-banner",
+      "value": true,
+      "reason": "TARGETING_MATCH",
+      "variant": "enabled"
+    }
+  ],
+  "refreshConnections": [
+    {
+      "type": "sse",
+      "url": "https://sse.example.com/event-stream?channels=env_abc123_v1",
```
JamieSinn, line 52: The URL should be split between `hostname` and `path` properties.

This allows for systems that need to override the endpoint without having to hackily substring and replace.  Generally required by larger/proxied users.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: I agree with todd on the naming here

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -0,0 +1,291 @@
...
+- `inactivityDelaySec` (integer, optional): Seconds of client inactivity (e.g., browser tab hidden, mobile app backgrounded) after which the connection should be closed. The client must reconnect and perform a full unconditional re-fetch when activity resumes. Minimum value is `1`. If omitted, providers should default to `120` seconds.
+
+The `refreshConnections` field is an array to support vendors whose infrastructure may require connections to multiple channels or endpoints (e.g., a global channel for environment-wide changes and a user-specific channel for targeted updates). Many SSE providers support multiple channels on a single URL, so the array will typically contain a single entry.
+
+### SSE Event Format
+
+Events use the standard [SSE event format](https://html.spec.whatwg.org/multipage/server-sent-events.html) with a JSON `data` field:
```
JamieSinn: Note - the `data` property is a string/byte array, not a discrete object due to the handling of the multiple `data` properties in the request resulting in concatenation. 

This is how we currently do this in DevCycle; and parse the JSON from this byte array.
https://github.com/DevCycleHQ/go-server-sdk/blob/main/ssemanager.go#L148

toddbaert (approved):
I have mostly nits, small additions, and naming improvement suggestions. Overall I think this is a great addition. 

I do agree with @lukas-reining 's point here: https://github.com/open-feature/protocol/pull/63#discussion_r2873456755

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: @toddbaert. I agree — server-side use cases are valid for SSE change notifications.

I think `refreshConnections` should be included in the individual flag eval response too, not just bulk eval. That covers server-side providers using the single-flag endpoints.

The remaining gap is local evaluation — providers doing in-process evaluation aren't calling OFREP eval endpoints at all, so there's no response to carry `refreshConnections`. A standalone endpoint like `GET /ofrep/v1/refreshConnections` could serve that case, but I'd defer that to a follow-up ADR.

I'll update the ADR to remove the static-context-only scope and include `refreshConnections` in the single-flag eval response schema.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: We'll try to update to OpenAPI 3.2.0 and add the `text/event-stream` representation.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: Makes sense, I'll rename `refreshConnections` → `eventStreams` and `refreshConnection` → `eventStream`.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: Makes sense, I'll rename `sseEtag` → `flagConfigEtag` and `sseLastModified` → `flagConfigLastModified`.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: I think keeping `url` as a single opaque string is the right call here. The server fully controls the URLs it returns in the response, so if a proxied setup needs different endpoints the server can return the correct URLs directly. I don't think we should be modifying these URLs in the provider layer.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: Yea good clarification — I'll add a note that the SSE `data` field is a raw string per the W3C spec and providers must parse it as JSON. The current wording could imply it's natively a JSON object.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: Yea agreed, that's a good way to frame the distinction. I'll add that rationale to the open question answer.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: Happy to book a discussion on this next week if you want @lukas-reining

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: > I think keeping url as a single opaque string is the right call here.

I think so for the same reason.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: > The remaining gap is local evaluation — providers doing in-process evaluation aren't calling OFREP eval endpoints at all.

But this isn't something supported by OFREP anyways. Or am I getting you wrong? 

> A standalone endpoint like GET /ofrep/v1/refreshConnections could serve that case, but I'd defer that to a follow-up ADR.

To me this feels like the way we should implement this change for the servers too. 
If we had the diff here as event payload it would make even more sense. 

My feeling is that a SSE connection for each flag would be way to much in most cases.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: Sounds good!

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: > Happy to book a discussion on this next week if you want @lukas-reining

Yeah I think that would be the easiest! Let's schedule something @jonathannorris  :)

lukas-reining: > I see your concerns about "thundering herd" and it totally makes sense in some cases, but due to the decentralized architecture we are working on it's less of a realistic problem for us...

Yeah I also totally understand that this is not a problem in the architecture Jonathan and you describe. 
But to me this feels like something only done in SaaS or very large deployments. 

From my experience with customers, many go for a single flagd or similar instance hosted in their typical environment. And there the "thundering herd" problem is real with a larger amount of clients. 
And they have good reasons and constraints to go this route. So actually in my "bubble" this is the much more common case. 

There having event pushed to an already open SSE connection is much more efficient. 

>  It also seems to me that things like jitter, etc could help here, and that's probably more of an implementation than an API question.

This is true! But I feel like these implementation level concerns should only be the last resort if the system can not be done "better".

lukas-reining: > Otherwise, maybe you can just see if the extensibility @jonathannorris has proposed is sufficient and do a follow up (I'm happy to help with this).

The extensibility is sufficient and I am happy to do this in a follow-up PR then. @toddbaert @jonathannorris 
I must admit that I will not be able to do it before KubeCon :D 
But after that I can work on an extension to support "change sets" or something.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: A proxied setup will need to be able to re-write the URL; while keeping the path and auth/channel specifics it will need to be able to re-write the hostname and protocol. 

Using a regex solution to do this in the proxy doesn't exactly seem the correct solution, and quite hacky.

jonathannorris: Made an update here to support proxied/origin-rewrite setups without changing the default model.

`url` remains the default representation, but the ADR now also allows a structured `endpoint` form with `origin` + `requestUri` for implementations that need to override the origin cleanly while preserving the request target. The intent is that exactly one of `url` or `endpoint` is provided for an event stream entry.

jonathannorris:
@lukas-reining / @thomaspoignant any further questions or approvals before I merge this?

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: I am not sure about this. 
How would that work? How would a provider implementation use the `origin + requestUri` to support a proxied setup? 
@JamieSinn could you elaborate on how you would imagine to use this? 
I would like to avoid any extra options that are not necessary or giving a larger benefit.

lukas-reining (approved):
Approval from me, I would just like to have sorted out the proxy url idea before merging. 
Thanks again @jonathannorris, I know this took a lot of time :)

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: This is something we're likely to use internally for server-side SSE connections at Dynatrace.

In our previous proxy setup at DevCycle, we effectively re-broadcast SSE messages from the proxy to internal SDKs/providers. In that model, the provider needs to connect to the local proxy origin, while preserving the path/query/channel-specific parts of the original SSE endpoint. Keeping this as a single opaque url makes that override logic pretty awkward and usually turns into string replacement.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
jonathannorris: yea, local evaluation isn't covered by OFREP today.

What I meant is that the same functionality is also needed for local evaluation providers, like providers using flagd in-process evaluation.

At DevCycle we use the same SSE config update mechanism for both remote evaluation and local evaluation. For local in-process evaluation SDKs, the SSE connection details are just included in the CDN config, and the SSE event tells the SDK to fetch updated config values from the CDN.

Long-term, it may make more sense to model that as an OFREP endpoint instead, but I’d probably leave that as an open question for now rather than trying to solve both shapes in this ADR.

I also agree I wouldn't want one SSE connection per flag. The event semantics should still just be “the underlying flag configuration changed”.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
lukas-reining: > This is something we're likely to use internally for server-side SSE connections at Dynatrace.

Yesh I guessed so. 

> In that model, the provider needs to connect to the local proxy origin, while preserving the path/query/channel-specific parts of the original SSE endpoint. Keeping this as a single opaque url makes that override logic pretty awkward and usually turns into string replacement.

I am not convinced of that. In my view adding special parameters for a non-standard case that can trivially be replaced by leaving the splitting of the URL to the implementer of that proxy logic.

JS Example: 

```js
const urlString = 'https://example.com:8080/path/to/page?param1=value1&param2=value2#hash';
const url = new URL(urlString);

const origin = url.origin;
const pathAndParams = url.pathname + url.search + url.hash; 
```

JamieSinn: Yep - happy to expand on it. 

With DevCycle - we had to split this between hostname and path due to the way that our downstream SDK proxy needed to split the endpoints and rewrite them based on the host request parameters that were incoming.
You can see some of this implementation here:
https://github.com/DevCycleHQ/sdk-proxy/blob/main/http_endpoints.go#L247
https://github.com/DevCycleHQ/sdk-proxy/blob/main/http_endpoints.go#L199

If we were to implement the url concept and the SDK proxy were to operate in cloud mode (which it can, and quite a few customers do use it in) - then there is no way to re-write the url property without putting the breaking nature of the path/upstream origin on the SDK proxy implementation. 

This is definitely an ease of use request for the proxy use case - but I don't really see the harm in the additional option to have your hostname/endpoint path if the default is the full URL.

PR #64, opened 2026-03-06
docs: add ADR 0009 for static-context cache-first local persistence

### LINKED ISSUE #65: OFREP Static-Context Provider Local Persistence
## Problem

OFREP static-context providers currently use an in-memory bulk evaluation cache. This works while the OFREP service is reachable, but the cache is lost across application restarts and cannot be reused during offline startup or temporary network loss.

This is a poor fit for web and mobile environments, where intermittent connectivity and app restarts are common. A client may have a previously successful bulk evaluation, but without local persistence it falls back to errors or code defaults instead of continuing from the last known state.

## Proposed Solution

Recommend that static-context providers persist their last successful bulk evaluation in local persistent storage by default.

A provider could store a single local cache record containing the bulk evaluation payload, the associated `ETag`, a derived cache-key hash based on inputs such as the `targetingKey` and auth token, and the time the entry was written.

On initialization, the provider would load the persisted entry, attempt the normal bulk evaluation request, and fall back to the persisted entry only when the request cannot complete because the client is offline or the network is temporarily unavailable.

Further details and proposed provider behavior will follow in an ADR PR.

jonathannorris:
## Summary
- add ADR 0009 proposing local persistence with cache-first initialization for static-context OFREP providers
- on startup, providers load cached evaluations immediately so initial flag evaluations never return defaults, then refresh from the network in parallel
- define cache-hit and cache-miss initialization paths mapped to the OpenFeature provider lifecycle (`PROVIDER_READY`, `PROVIDER_CONFIGURATION_CHANGED`)
- cache key uses `hash(targetingKey)`, no auth token dependency

## Motivation
Every major vendor SDK (LaunchDarkly, Statsig, DevCycle, Eppo) uses cache-first initialization by default. Current OFREP static-context providers keep their cache in memory only, losing all state on restart. See [vendor mobile SDK caching research](https://gist.github.com/jonathannorris/4f2f63142b70719e3c6bfe8b226a0585) for a detailed comparison.

## Related
- Closes #65

## Test plan
- not applicable; documentation-only change


service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Decision
+
+Static-context providers should persist their last successful bulk evaluation in local persistent storage by default.
+
+The persisted entry should include:
+
+- the bulk evaluation payload
+- the associated `ETag`, if one was returned
+- a `cacheKeyHash` equal to `sha256(authToken + targetingKey)`
```
beeme1mr: Shouldn't we hash the full evaluation context?

lukas-reining: I think so too. 
`targetingKey` is never granted to be the context that produced the evaluation.

lukas-reining: I would not specify `sha256(authToken + targetingKey)` here. 
We can do it in the provider guideline but to me this should not be specified across platforms or providers as this is basically an implementation detail.

lukas-reining: What is the `authToken` for? 
In my view, the `authToken` could be refreshed multiple times while the context and even content stay fresh. Also: There is not always an auth token. 
As @beeme1mr said, to me the only relevant cache key should be the context.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+- the time the entry was written, which can be used for diagnostics and optional implementation-specific staleness policies
+
+Providers may store this as a single fixed local record, for example under a runtime-appropriate key such as `ofrepLocalCache`, and replace that record on each successful refresh.
```
lukas-reining: Why do we specify this? 
To me this is an implementation detail.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+The provider should continue to use its in-memory cache for normal flag evaluation.
+Persistent local storage acts as the source used to bootstrap or recover that in-memory cache.
+
+During initialization, a provider should:
+
+1. Attempt to load a matching persisted bulk evaluation from local storage.
+2. Attempt the normal `/ofrep/v1/evaluate/flags` request.
+3. If the request succeeds, populate the in-memory cache from the response and update the persisted entry.
+4. If the request cannot complete because the client is offline, the network is temporarily unavailable, or the server is temporarily unavailable, such as a `5xx` response:
+   - If a matching persisted entry exists, populate the in-memory cache from that persisted entry and continue operating from it.
```
lukas-reining: Maybe we can specify that `CACHED` should be the evaluation reason then? 
Or shouldn't it because we want it to be transparent? 

https://openfeature.dev/specification/sections/providers#requirement-225

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+    else Flags unchanged
+        Server-->>Provider: 304 Not Modified
+    end
+```
+
+Providers should only reuse a persisted evaluation when it matches the current static-context inputs.
+This includes a matching `cacheKeyHash` equal to `sha256(authToken + targetingKey)`.
+
+Fallback to persisted data is intended for offline, transient network failures, or temporary server unavailability such as `5xx` responses.
+Providers should not silently fall back to persisted data for authorization failures, invalid requests, or other responses that indicate a configuration or protocol problem.
```
lukas-reining: I know this is not formal language but to me we should go for `may not` here.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+- The decision aligns with the existing OFREP model where static-context providers evaluate remotely once and then read locally
+- Reusing the stored `ETag` allows efficient revalidation when connectivity returns
+- Provider implementations get a consistent default expectation for offline behavior across ecosystems
+
+### Negative
+
+- Providers become more complex because they must manage persistence, cache-key matching, and recovery flows
+- Persisted evaluations may become stale, so applications can continue using outdated flag values while offline
+- Local persistent storage can be unavailable, limited in size, or restricted by platform policy
+- Persisting evaluation data introduces security and privacy considerations, especially if flag metadata or context-derived identifiers are sensitive
```
lukas-reining: `introduces security and privacy considerations` can we make this a little more concrete?

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+- Web and mobile applications preserve feature state across restarts instead of losing it with the in-memory cache
+- The decision aligns with the existing OFREP model where static-context providers evaluate remotely once and then read locally
+- Reusing the stored `ETag` allows efficient revalidation when connectivity returns
+- Provider implementations get a consistent default expectation for offline behavior across ecosystems
+
+### Negative
+
+- Providers become more complex because they must manage persistence, cache-key matching, and recovery flows
+- Persisted evaluations may become stale, so applications can continue using outdated flag values while offline
+- Local persistent storage can be unavailable, limited in size, or restricted by platform policy
```
lukas-reining: I think this is not a negative consequence of the proposal.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+
+This reduces default behavior changes, but it produces inconsistent offline behavior across provider implementations and requires every application to rediscover and enable the same capability.
+For static-context providers, especially web and mobile providers, persistence is expected behavior rather than an exceptional optimization.
+
+## Implementation Notes
+
+- "Local storage" means a local persistent key-value store appropriate for the runtime, such as browser `localStorage` on the web or an equivalent mobile storage mechanism
+- Providers should version their persisted format so future schema changes can be handled safely
+- Providers may use a single fixed storage key or filename and store the matching information inside the record as `cacheKeyHash`
+- `cacheKeyHash` should be `sha256(authToken + targetingKey)`
```
lukas-reining: Again, I am not sure if this is a good idea. 
The `authToken` should be completely independent from this in my view and for the static-context paradigm, the whole context is necessary to determine the evaluated flag values. 

Also, I would not like to specify `sha256`. There are reasons to have this specified, but to me leaving this as implementation detail pays of more to me.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+- Providers should avoid persisting raw `targetingKey` and auth token values when `cacheKeyHash` is sufficient for matching
+- Providers should expose a `disableLocalCache` option to turn off persisted local storage
+- Providers should clear or replace persisted entries when the `targetingKey` or auth token changes, such as on logout or user switch
+- SDK documentation should describe that offline fallback uses the last successful bulk evaluation and may therefore serve stale values until connectivity returns
+
+## Open Questions
+
+1. Should providers fall back to persisted data only when the client is offline or the network is temporarily unavailable, or should they also fall back for authorization failures, invalid requests, or other server responses that indicate a configuration or protocol problem?
```
lukas-reining: To me, it feels like all transient and non client induced errors could make sense here. 
If the ff backend returns a 503 for some reason, I think falling back to the cache makes sense. 
This is also in line what happens with the in-memory cache.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+2. Should providers also persist the full evaluation context used for the cached bulk evaluation, so that when falling back to persisted values they can override the current context with the cached context that produced those values?
```
lukas-reining: I think this is a very good idea! 
As said above, I think caching context only makes sense in the context of the full evaluation context. 
So if the app is started without a network connection it should not simply get the last evaluation results independent from the evaluation context. 
If we go for the last evaluation context and use the full context as cache key, we are back in a consistent spot. 
The question here would be, how the provider could set the static context in the SDK.

lukas-reining (commented):
Thanks @jonathannorris, left some feedback. 
As you, I think this is a very important addition to the OFREP providers. 

I have some thoughts regarding the `cache key`, which could also influence the whole concept due to the static context possibly also not being available as you said in open question no. 2.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,151 @@
...
+      }
+    ]
+  }
+}
+```
+
+The provider should continue to use its in-memory cache for normal flag evaluation.
+Persistent local storage acts as the source used to bootstrap or recover that in-memory cache.
+
+During initialization, a provider should:
```
guidobrei: Couldn't we achieve the desired behavior with standard HTTP Cache directives?

Sketching the flow:

1. Client generates a `cacheKey` hash based on context and other custom data required to determine distinguishable requests as described above, *with the difference, that it is client generated not server generated*.
2.  Client attempt evaluation request to `POST /ofrep/v1/evaluate/flags?cacheKey=<cacheKey>`. *Appending cacheKey as query pram makes request URLs unique and cacheable*
3. OFREP server responds with `Cache-Control` and `Etag` header
```
ETag: "abc123"
Cache-Control: private, max-age=0, stale-if-error=86400
```
*`private` allows local caching, `max-age=0` requires revalidation, `stale-if-error` allows the cache to serve stale content for some time if offline or on defined 5xx errors.*

Things to consider (have not investigated much):
* Do browsers/mobile HTTP caches honor cache directives on `POST` request? 
* Are Proxies in line rewriting Response headers?
* Is `stale-if-error` Cache-Control extension ([rfc5861](https://www.rfc-editor.org/rfc/rfc5861)) supported by all target platforms?
* The approach does not fall back to cached content on 401/403 responses

lukas-reining: I would really like to just use HTTP caching but to me the considerations you are mentioning weigh in pretty much. 

> Do browsers/mobile HTTP caches honor cache directives on POST request?
 
[RFC7234 4.3.3](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.3): 

```
Responses to POST requests are only cacheable when they include
explicit freshness information (see Section 4.2.1 of [RFC7234]).
However, POST caching is not widely implemented. For cases where an origin server wishes the client to be able to cache the result of a POST in a way that can be reused by a later GET, the origin server MAY send a 200 (OK) response containing the result and a Content-Location header field that has the same value as the POST's effective request URI (Section 3.1.4.2).
```

> The approach does not fall back to cached content on 401/403 responses

This and the general dependence on client, proxies and and other parts of the system makes this possibly harder than it seems.

guidobrei: Agree, but even if we implement cache in provider land, we could make use standard HTTP headers, instead of packing the fields into the content.

Handling 401/403 is also challenging in @jonathannorris approach, as the cacheKey is server generated, and auth error responses are likely generated by some gateway before the request even hits the OFREP server.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: @beeme1mr I think hashing the full evaluation context is problematic here. Many implementations set time-based or session-based context properties on startup (e.g. `lastSessionTime`, `lastSeen`, `sessionId`), so hashing the full context would mean the cache key changes on every app restart, which defeats the purpose of persisting it for offline bootstrap.

I think it probably makes sense to default the cache key to `hash(targetingKey)` since for static-context providers the targeting key is the primary discriminator for "is this cached data relevant to the current user?" If the context has changed in other ways, those differences only matter when the server is reachable, and at that point the provider refreshes anyway.

Open to other ideas here, but I think full context hashing would break the offline use case for a lot of real-world implementations.

jonathannorris: yea agree, will switch to a generic `hash()` and leave the algorithm as an implementation detail.

jonathannorris: @lukas-reining yea `authToken` should be dropped from the cache key entirely. My original thinking was for situations where the project you're targeting against could potentially change, but that seems pretty unlikely in the real world. And systems using OAuth refresh tokens would invalidate the cache on every token rotation even though the evaluation context hasn't changed. Will remove it.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: Addressed in the line 28 threads. Dropping `authToken`, switching to generic `hash()`, and using `targetingKey` as the default cache key input.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: I think setting the full evaluation context on the SDK from the provider is probably outside the scope of what OFREP providers should be doing. That would fundamentally change how the SDKs behave.

Using `hash(targetingKey)` as the cache key handles the primary case: user switch or logout produces a different targeting key, so the old cache entry won't match.

One thing I'm wondering though is whether we should consider a more aggressive pattern here. The current ADR only loads from cache on network failure, but in our SDKs at DevCycle we load the last cached evaluations and user data immediately on startup so initial synchronous flag evaluations never return defaults, then refresh from the network in parallel and fire a change event when fresh values arrive. The UX is meaningfully better since you don't flash between defaults and evaluated values.

Should we consider expanding this ADR to cover that pattern instead of just cache-on-failure?

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: @guidobrei interesting idea, but HTTP POST response caching is not supported on any of the major mobile or web clients. iOS `NSURLCache` only caches GET responses, OkHttp and `HttpURLConnection` on Android only cache GET responses, and the Fetch API in browsers skips cache for non-GET methods per spec. `stale-if-error` (RFC 5861) is also not supported by any of these clients, it's primarily a CDN/proxy-level feature.

With provider-land caching we have full control over when we write, when we read, and what the fallback logic looks like. I think that's necessary for the offline bootstrap use case.

re: the server-generated cacheKey concern, that's addressed now since we've switched to `hash(targetingKey)` which is fully client-derived.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: yea I agree, falling back on all transient and non-client-induced errors makes sense. The decision section already takes this position (lines 62-65), so I think we can just remove this open question.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: yea I think we should use `CACHED` as the evaluation reason when serving from persisted storage. It's transparent to the consumer and aligns with the spec's defined reasons. Will update the ADR.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: I think this should actually be stronger. Updated to `must not` since falling back to cached data on auth or config errors would mask real problems.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: yea fair point, that's a platform constraint not a consequence of the proposal. Removed.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: Made this more concrete. Flag values are stored in plaintext in platform-local storage, which may be accessible to other code on the same origin or on compromised devices.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: yea agree, removed. The storage model is an implementation detail.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: Based on some AI-assisted research, it seems like most vendors bootstrap values from the local cache before the network call. I think we should probably re-work this ADR with that approach: https://gist.github.com/jonathannorris/4f2f63142b70719e3c6bfe8b226a0585

lukas-reining: > Should we consider expanding this ADR to cover that pattern instead of just cache-on-failure?

Yeah that's what I also meant in the other comment. To me that sounds sensible!

lukas-reining: > I think setting the full evaluation context on the SDK from the provider is probably outside the scope of what OFREP providers should be doing. That would fundamentally change how the SDKs behave.

Yeah that was my concern too. 
I guess targeting key + offline can be okay considering that targeting rules are meant to be transparent to the app author in most cases, meaning that even changing context (not targetingId) in an offline scenario would be okay and not lead to much confusion.

service/adrs/0009-localStorageForStaticContextProviders.md
lukas-reining: > My original thinking was for situations where the project you're targeting against could potentially change, but that seems pretty unlikely in the real world.

Yeah I think this is the same argumentation as for the rest of the context besides the targeting key. 

> And systems using OAuth refresh tokens would invalidate the cache on every token rotation even though the evaluation context hasn't changed.

Exactly this is why I think auth token is not suitable.

lukas-reining: I think the hard part about this is making this clear to the app author. 
Your argumentation makes sense to me. And I think it is okay to use the special semantics of the targeting key for this. 
Even though it could be confusing that changing any other context does not change the evaluations in an offline scenario, I think it is acceptable as caching is `half-transparent` to the app author as they can always see the evaluation reason.
We should still make a good job documenting this as for outside standing people this might not be obvious?

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: Update: the ADR has been rewritten to use cache-first initialization. The provider now loads from the persisted cache immediately on startup and returns from `initialize()` so the SDK emits `PROVIDER_READY`. The network refresh happens in the background and emits `PROVIDER_CONFIGURATION_CHANGED` when fresh values arrive. This aligns with how vendor SDKs (LaunchDarkly, Statsig, DevCycle, Eppo) handle it. See the [vendor research](https://gist.github.com/jonathannorris/4f2f63142b70719e3c6bfe8b226a0585) for details.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,208 @@
...
+### Initialization
+
+During initialization, a provider should follow a cache-first approach:
+
+1. Attempt to load a matching persisted bulk evaluation from local storage (matching `cacheKeyHash`).
+2. **If a matching persisted entry exists (cache hit):**
+   - Populate the in-memory cache from the persisted entry immediately.
+   - Return from `initialize()` so the SDK can emit `PROVIDER_READY`. Evaluations served from the persisted entry should use `CACHED` as the evaluation reason.
+   - Attempt the `/ofrep/v1/evaluate/flags` request in the background.
+   - If the background request succeeds, update the in-memory cache from the response, update the persisted entry, and emit `PROVIDER_CONFIGURATION_CHANGED`. Evaluations should switch to the server-provided reasons.
```
toddbaert, line 68: The cache-hit returns from initialize() immediately and fires a background refresh... IIUC, this means an app could evaluate a flag, get the cached value, then receive `PROVIDER_CONFIGURATION_CHANGED` with a different value milliseconds later. This could result in flickering and UI thrashing. This is a direct tradeoff of the cache-first design. I think this is mentioned in the negatives, but I'm interested to know what, if anything, can be done to improve this.

Are we sure that most users would consider loading these cached values first, and potentially seeing a change immediately at app load, better than seeing some loaders/spinners while we attempt to evaluate flags the first time?

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Initialization
+
+During initialization, a provider should follow a cache-first approach:
+
+1. Attempt to load a matching persisted bulk evaluation from local storage (matching `cacheKeyHash`).
+2. **If a matching persisted entry exists (cache hit):**
+   - Populate the in-memory cache from the persisted entry immediately.
+   - Return from `initialize()` so the SDK can emit `PROVIDER_READY`. Evaluations served from the persisted entry should use `CACHED` as the evaluation reason.
+   - Attempt the `/ofrep/v1/evaluate/flags` request in the background.
```
toddbaert, line 67: Is it worth thinking about what happens if `setContext()` is called while this "background" refresh is still in-flight? Might be good to add a note that providers should serialize writes to the persisted entry to avoid weird race conditions or states resulting from out-of-order responses?

Maybe it's overkill, but just an idea.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: @toddbaert I think the reality of this is in practice that 99% of flags don't change between the last cached value and the updated bulk evaluation. So yes there is a chance that users see a thrash in UI, but the trade-off for not blocking rendering and creating a poor experience in low-network situations is much greater than that risk.

jonathannorris: Also if there is no local cache, the `initialize()` function does block on a bulk OFREP request. Maybe there is a case to be made for having the TTL on the cache be lower?

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: Good catch. Providers should cancel any in-flight background refresh when `onContextChanged()` is called — the context-change evaluation supersedes it. Will add a note to the implementation notes.

service/adrs/0009-localStorageForStaticContextProviders.md
toddbaert: Ya, that makes sense. Just wanted to make sure.

toddbaert (approved):

jonathannorris:
@lukas-reining / @thomaspoignant take a look at this ADR when you get a chance

service/openapi.yaml
```diff
@@ -1,8 +1,8 @@
-openapi: 3.1.0
+openapi: 3.2.0
```
lukas-reining: Same comments regarding Webhook and OpenAPI 3.2 as in https://github.com/open-feature/protocol/pull/67#pullrequestreview-4046705620 apply.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,211 @@
...
+   - If the background request succeeds, update the in-memory cache from the response, update the persisted entry, and emit `PROVIDER_CONFIGURATION_CHANGED`. Evaluations should switch to the server-provided reasons.
+   - If the background request fails with a transient or server error (network unavailable, `5xx`), continue serving cached values and retry on the normal polling schedule.
+   - If the background request fails with an authorization or configuration error (`401`, `403`, `400`), surface the error via logging or provider error events, continue serving cached values for the current session, and clear the persisted entry from local storage. This ensures the next cold start uses the cache-miss path, making the auth or configuration error immediately visible rather than silently booting from increasingly stale data.
```
lukas-reining, line 70: I really like this and I think it makes sense for the general case. 
I am wondering if it could make sense to make the clearing optional with the default being enabled. 
Especially for 400 errors I could imagine that you could want the old values not be cleared in some cases, while I agree that clearing should be the default.

lukas-reining: I am going back and forth on this. Clearing seems very sensible. On the other hand, `PROVIDER_ERROR` then always comes on the `n+1` request after the errored request.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,211 @@
...
+
+The spec defines `READY` as "the provider has been initialized, and is able to reliably resolve flag values" and `STALE` as "the provider's cached state is no longer valid and may not be up-to-date with the source of truth."
+
+On cache-hit startup, the provider emits `PROVIDER_READY` rather than `PROVIDER_STALE` for two reasons.
+First, at the moment of loading from cache, the provider does not yet know whether the cached values differ from the server. The values were correct as of the last successful evaluation and may still be current. The background refresh will determine whether they have changed.
+Second, `PROVIDER_STALE` would break the initialization contract. Applications and SDKs listen for `PROVIDER_READY` to begin flag evaluation. If the provider emitted `PROVIDER_STALE` instead, the SDK would not transition out of `NOT_READY`, and flag evaluations would short-circuit to defaults, which defeats the purpose of cache-first initialization.
+
+If the background refresh fails and the provider cannot confirm that cached values are current, the provider may emit `PROVIDER_STALE` at that point to signal that values may be out of date.
```
lukas-reining, line 137: I think this is a good compromise.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Cache matching and fallback
+
+Providers should only reuse a persisted evaluation when it matches the current static-context inputs.
+This includes a matching `cacheKeyHash` equal to `hash(targetingKey)`.
+
+The cache key is intentionally derived from `targetingKey` alone rather than the full evaluation context.
+Static-context evaluations on the server can depend on context properties beyond `targetingKey`, so cached values may not reflect the current full context.
+However, hashing the full context is impractical for cache-first startup because many implementations set volatile context properties on initialization (e.g. `lastSessionTime`, `lastSeen`, `sessionId`) that would change the hash on every app restart, defeating the purpose of persistence.
```
lukas-reining, line 146: Do you think it makes sense to allow additional cache key properties to be configured in the provider constructor? This would allow us to support different use-cases without having to change the spec later. 
Or are we waiting until someone comes up with the need?

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Decision
+
+Static-context providers should persist their last successful bulk evaluation in local persistent storage by default, and use cache-first initialization to serve persisted evaluations immediately on startup.
+
+The persisted entry should include:
+
+- the bulk evaluation payload
+- the associated `ETag`, if one was returned
+- a `cacheKeyHash` equal to `hash(targetingKey)`
```
lukas-reining, line 31: I think we should include the OFREP API URL in this. 
We use multiple OFREP servers in our services using the same `targetingKey`.
Without this, the second provider would just go for the cache from the first server.

I think `ofrepUrl + targetingKey` would be okay, as this identifies the server dependent flag cache. 
The only problem is if the providers would only differ in something like a `tenant: abc` HTTP header, because this would not be considered here. 

Alternatively I thought we could go for the `domain + targetingKey`, meaning the OpenFeature domain. 
If we did this, the domain would have to be passed to the provider on initialize. This should even work if a specific domain uses the default provider as the key `default` would still be correct. 
But I am not sure if we want to do the SDK change for this.

Yet another option could be to just leave it up to the app author to set different cache prefixes in the provider if needed. But this seems like the worst option to me as app authors would probably have to run into problems to discover this.

service/adrs/0009-localStorageForStaticContextProviders.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Open Questions
+
+1. Should providers support caching evaluations for multiple targeting keys (like LaunchDarkly's `maxCachedContexts`), or only retain the most recent? Multi-context caching enables instant user switching on shared devices but increases storage usage.
+2. Should providers enforce a TTL on persisted entries (e.g. 30 days, similar to DevCycle's `configCacheTTL`)? A TTL would ensure stale caches are eventually purged, particularly in cases where the provider can no longer refresh from the server (e.g. persistent auth errors). If so, should the TTL be configurable?
+3. Should the storage key include a namespace to prevent collisions when multiple OFREP providers share the same local storage origin (e.g. different backends on the same web origin)? In practice most applications use a single provider, so real-world collisions are unlikely. One option is to namespace using a hash of the auth token, since it is already environment- and project-specific and effectively distinguishes one provider configuration from another.
```
lukas-reining: As described here https://github.com/open-feature/protocol/pull/64/changes#r3024572818 I really think we should support more than one OFREP provider instance.

lukas-reining (commented):
This looks good to me all in all! 
Left some questions/thoughts.

service/openapi.yaml
jonathannorris: yea this has been addressed in #67 — the webhook was moved to `components/pathItems/eventStreamConnection` and the spec was updated to use the proper path item component approach. That PR also handles the other OpenAPI structure feedback. Once #67 lands this will be resolved here.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: @lukas-reining yea after looking into how vendor SDKs handle this (e.g. DevCycle keeps the cache through auth errors and relies on a configurable TTL for eventual expiry), I've updated the ADR to match that approach. The persisted cache is no longer cleared on auth/config errors. The TTL is responsible for eventual expiry and will be configurable. This avoids the n+1 problem you flagged.

See the updated ADR in #75: https://github.com/open-feature/protocol/pull/75/commits/76ac0bc

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: @lukas-reining I think this ADR already adds a good amount of complexity to the provider implementations. My instinct is to keep configurability minimal for now and add it when there's a concrete need.

Continued in #75.

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: @lukas-reining rather than baking `ofrepUrl` into the spec, I think a configurable `cacheKeyPrefix` option on the provider constructor is the right approach. When provided, the cache key becomes `hash(cacheKeyPrefix + targetingKey)`. This covers multiple use cases without the spec prescribing what the prefix should be: it could be the base URL, a project/auth token, or any other string. The default (no prefix) keeps the single-provider case simple.

Updated the open question in #75: https://github.com/open-feature/protocol/pull/75/commits/6b80b3e

service/adrs/0009-localStorageForStaticContextProviders.md
jonathannorris: Addressed in the [cache key namespace thread](https://github.com/open-feature/protocol/pull/64#discussion_r3075409817) — proposing a configurable `cacheKeyPrefix` option.

PR #67, opened 2026-03-16
feat: implement ADR-0008 event stream support in OpenAPI spec

jonathannorris:
## Summary

- Implements the OpenAPI spec changes defined in [ADR-0008](service/adrs/0008-sse-for-bulk-evaluation-changes.md), scoped to static-context providers using the bulk evaluation endpoint (`POST /ofrep/v1/evaluate/flags`)
- `eventStreams` is intentionally limited to `bulkEvaluationSuccess` — SSE support for dynamic-context providers using single-flag evaluations is deferred to a follow-up ADR
- Bumps OFREP version from 0.2.0 to 0.3.0
- Adds `eventStreams` to `bulkEvaluationSuccess`, `flagConfigEtag` / `flagConfigLastModified` query parameters to the bulk eval endpoint, and new schemas for event stream connections and event payloads
- Adds a separate `event-streams.yaml` (OAS 3.2.0) documenting the `text/event-stream` + `itemSchema` event format, while keeping the main `openapi.yaml` at OAS 3.1.0 with full spectral linting

### Notes

- The `eventStream` schema enforces mutual exclusivity between `url` and `endpoint` using `oneOf` + `not` clauses
- `endpoint` provides a structured `requestUri` (required) and optional `origin` for proxied deployments that need to override the origin while preserving the request target; if `origin` is absent, providers use their configured OFREP base URL
- Query parameter descriptions use generic "event stream" / "change notification" language rather than SSE-specific terms, since `eventStreams` is designed to be extensible beyond SSE
- The main spec stays at OAS 3.1.0 with `oas3-schema` linting re-enabled; the OAS 3.2.0 `text/event-stream` + `itemSchema` representation lives in `service/event-streams.yaml`

### Related Issues

- Implements: ADR-0008
- Relates to: #63

### DIFF
--- a/service/adrs/0008-sse-for-bulk-evaluation-changes.md
+++ b/service/adrs/0008-sse-for-bulk-evaluation-changes.md
@@ -62,10 +62,10 @@ Add an optional `eventStreams` field to `bulkEvaluationSuccess`:
 Each event stream object has:
 - `type` (string, required): The connection type. Currently `"sse"` is the only defined value. Providers must ignore entries with unknown types for forward compatibility, allowing new push mechanisms to be added without breaking existing clients.
 - `url` (string, optional): The endpoint URL. This is the default representation and is opaque to the provider. It may include authentication tokens, channel identifiers, or other vendor-specific query parameters. Implementations must treat this URL as sensitive -- it may contain auth tokens or channel credentials -- and must not log or persist the full URL including query string.
-- `endpoint` (object, optional): Structured endpoint components for deployments that need to override the origin cleanly (for example, via a proxy) while preserving the request target. If present, it has `origin` and `requestUri` fields.
+- `endpoint` (object, optional): Structured endpoint components for deployments that need to override the origin cleanly (for example, via a proxy) while preserving the request target. It has a required `requestUri` field and an optional `origin` field. If `origin` is absent, providers should use their configured OFREP base URL origin.
 - `inactivityDelaySec` (integer, optional): Seconds of client inactivity (e.g., browser tab hidden, mobile app backgrounded) after which the connection should be closed. The client must reconnect and perform a full unconditional re-fetch when activity resumes. Minimum value is `1`. When determining the effective inactivity timeout, providers should use a client-side override if configured; otherwise use this value when present; otherwise default to `120` seconds.
 
-Exactly one of `url` or `endpoint` must be provided. Providers should use `url` as-is when present. When `endpoint` is present, providers should construct the connection URL as `origin + requestUri`.
+Exactly one of `url` or `endpoint` must be provided. Providers should use `url` as-is when present. When `endpoint` is present, providers should construct the connection URL as `origin + requestUri`, where `origin` defaults to the provider's configured OFREP base URL if not specified.
 
 The `eventStreams` field is an array to support vendors whose infrastructure may require connections to multiple channels or endpoints (e.g., a global channel for environment-wide changes and a user-specific channel for targeted updates). Many SSE providers support multiple channels on a single URL, so the array will typically contain a single entry.
 
@@ -243,18 +243,20 @@ eventStream:
     endpoint:
       type: object
       required:
-        - origin
         - requestUri
       description: |
         Structured endpoint components for deployments that need to override
         the origin cleanly while preserving the request target. When present,
-        providers construct the connection URL as `origin + requestUri`.
+        providers construct the connection URL as `origin + requestUri`. If
+        `origin` is absent, providers should use their configured OFREP base
+        URL origin.
       properties:
         origin:
           type: string
           format: uri
           description: |
             The scheme + host + optional port portion of the endpoint URL.
+            If absent, providers should use their configured OFREP base URL origin.
           example: "https://sse.example.com"
         requestUri:
           type: string
--- a/service/adrs/0008-sse-for-bulk-evaluation-changes.md
+++ b/service/adrs/0008-sse-for-bulk-evaluation-changes.md
@@ -1,4 +1,4 @@
-# 8. Server-Sent Events (SSE) for bulk evaluation changes
+# 8. Server-Sent Events (SSE) for bulk evaluation changes (client-side providers)
 
 Date: 2026-02-20
 
@@ -10,7 +10,7 @@ Proposed
 
 OFREP currently relies exclusively on polling for flag change detection. As described in [ADR-0005](0005-polling-for-bulk-evaluation-changes.md), polling was chosen initially for simplicity, with the explicit expectation that additional change detection mechanisms would be added later.
 
-This ADR defines SSE as a real-time change notification mechanism for OFREP. The primary use case is static-context providers (web and mobile) that use bulk evaluation caching, but SSE is also applicable to server-side providers using individual flag evaluations. A standalone endpoint for providers doing in-process local evaluation (outside of OFREP) is deferred to a follow-up ADR.
+This ADR defines SSE as a real-time change notification mechanism for OFREP, scoped to client-side providers that use bulk evaluation caching. SSE support for server-side providers using individual flag evaluations and for providers doing in-process local evaluation (outside of OFREP) is deferred to a follow-up ADR.
 
 Polling has known limitations:
 - There is no way to implement real-time flag updates
@@ -28,13 +28,13 @@ Server-Sent Events (SSE) is a W3C standard that fits this use case well:
 
 ## Decision
 
-Add an optional `eventStreams` array to the bulk evaluation response (`POST /ofrep/v1/evaluate/flags`) and the single flag evaluation response (`POST /ofrep/v1/evaluate/flags/{key}`). When present, it provides connection endpoints that the provider connects to for real-time flag change notifications.
+Add an optional `eventStreams` array to the bulk evaluation response (`POST /ofrep/v1/evaluate/flags`). When present, it provides connection endpoints that the provider connects to for real-time flag change notifications.
 
 SSE is used as a **notification-only** mechanism -- events signal the provider to re-fetch evaluations via the existing endpoints, rather than streaming full evaluation payloads. This keeps the SSE message format simple, reuses existing infrastructure, and avoids duplicating evaluation logic.
 
 ### Response Schema
 
-Add an optional `eventStreams` field to `bulkEvaluationSuccess` and `serverEvaluationSuccess`:
+Add an optional `eventStreams` field to `bulkEvaluationSuccess`:
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


askpt (approved):

service/openapi.yaml
```diff
@@ -245,6 +260,43 @@ components:
...
+      description: |
+        Optional ETag metadata provided by an event stream for change-triggered
+        re-fetches (see ADR-0008). This is not a standard HTTP conditional request
+        header; it is metadata for server-side cache validation and freshness
+        checks. It should only be included when the request is directly triggered
+        by a received change notification event.
+      schema:
+        type: string
+      required: false
+      example: '"550e8400-e29b-41d4-a716-446655440000"'
```
beeme1mr: Should the example use both single and double quotes?

beeme1mr (approved):

service/openapi.yaml
```diff
@@ -79,6 +84,10 @@ paths:
                 value: true
                 reason: TARGETING_MATCH
                 variant: enabled
+                eventStreams:
```
erka: ADR-0008 says 'for bulk evaluation'. Is this out of the specification?

jonathannorris: Resolved in the latest commit — `serverEvaluationSuccess` no longer includes `eventStreams`. Event stream support is now scoped to the bulk evaluation endpoint only, based on ADR-0008 and discussions at KubeCon last week. SSE support for server-side providers is deferred to a follow-up ADR.

service/openapi.yaml
jonathannorris: The double quotes are intentional — ETags are quoted strings per [RFC 7232](https://datatracker.ietf.org/doc/html/rfc7232#section-2.3), so the value itself includes the surrounding double quotes. The outer single quotes are just YAML quoting. Happy to add a clarifying note to the description if that would help.

service/openapi.yaml
```diff
@@ -1,8 +1,8 @@
-openapi: 3.1.0
+openapi: 3.2.0
```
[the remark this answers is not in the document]
jonathannorris: https://spec.openapis.org/oas/v3.2.0.html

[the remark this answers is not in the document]
erka: 3.1.0 is a pain, I don't know about 3.2.0. GRPC supports only 3.0.0 out of the box.

service/openapi.yaml
```diff
@@ -308,6 +364,149 @@ components:
...
+            channel identifiers, or other query parameters as needed by the
+            vendor's infrastructure. Implementations must treat this value as
+            sensitive and must not log or persist the full URL including its
+            query string.
+          example: "https://sse.example.com/event-stream?channels=env_abc123_v1"
+        endpoint:
+          $ref: "#/components/schemas/eventStreamEndpoint"
+        inactivityDelaySec:
+          type: integer
+          minimum: 1
```
toddbaert, line 407: The prose says providers should default to 120s, but the schema does not declare it. Adding `default: 120` here would make this self-documenting for codegen tools.

```suggestion
          minimum: 1
          default: 120
```

toddbaert (commented):
Minor nit; looks great overall.

toddbaert (approved):
Approved, minor nit: https://github.com/open-feature/protocol/pull/67#discussion_r3017232299

jonathannorris:
@lukas-reining / @thomaspoignant / @erka any further comments on this one?

service/openapi.yaml
```diff
@@ -308,6 +417,160 @@ components:
...
+          description: |
+            The endpoint URL the client should connect to for real-time
+            flag change notifications. This is the default representation and
+            is opaque to the provider. The URL may include authentication tokens,
+            channel identifiers, or other query parameters as needed by the
+            vendor's infrastructure. Implementations should treat this value as
+            sensitive and should not log or persist the full URL including its
+            query string.
+          example: "https://sse.example.com/event-stream?channels=env_abc123_v1"
+        endpoint:
```
erka, line 403: @jonathannorris  Could you provide more information about this `endpoint` object and how it should be used for me? I don’t see a difference from `url` at the moment. While I understand the value of combining `endpoint.requestUri` with `OFREP_URL`, `endpoint.origin` seems to make it effectively the same as `url`. Is there something I might be missing?

jonathannorris: @erka here is the original discussion on it, let me know if you need any further details: https://github.com/open-feature/protocol/pull/63#discussion_r2931220752

erka: I support @lukas-reining's [statement](https://github.com/open-feature/protocol/pull/63#pullrequestreview-3955968466) - 'I would just like to have sorted out the proxy url idea before merging'.  

I’m not sure how much compute would be saved by this `origin` rewrite on the proxy. I’d prefer `endpoint.origin` to be optional rather than a required field.

lukas-reining: > I’m not sure how much compute would be saved by this origin rewrite on the proxy.

That's also my main reason.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -1,4 +1,4 @@
-# 8. Server-Sent Events (SSE) for bulk evaluation changes
+# 8. Server-Sent Events (SSE) for bulk evaluation changes — client-side providers
```
lukas-reining:
```suggestion
# 8. Server-Sent Events (SSE) for bulk evaluation changes — static-context providers
```

This would be more inline with our wording.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -10,7 +10,7 @@ Proposed
 
 OFREP currently relies exclusively on polling for flag change detection. As described in [ADR-0005](0005-polling-for-bulk-evaluation-changes.md), polling was chosen initially for simplicity, with the explicit expectation that additional change detection mechanisms would be added later.
 
-This ADR defines SSE as a real-time change notification mechanism for OFREP. The primary use case is static-context providers (web and mobile) that use bulk evaluation caching, but SSE is also applicable to server-side providers using individual flag evaluations. A standalone endpoint for providers doing in-process local evaluation (outside of OFREP) is deferred to a follow-up ADR.
+This ADR defines SSE as a real-time change notification mechanism for OFREP, scoped to client-side providers that use bulk evaluation caching. SSE support for server-side providers using individual flag evaluations and for providers doing in-process local evaluation (outside of OFREP) is deferred to a follow-up ADR.
```
lukas-reining: Why did you favor `scoped to client-side` over `static-context providers`?
To me `static-context` described more precise what behavior is expected.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -189,9 +189,9 @@ Provider implementation guidelines:
     isoDate:
       value: "2026-02-20T21:28:18Z"
     httpDate:
-      value: "Thu, 20 Feb 2026 21:28:18 GMT"
+      value: "Fri, 20 Feb 2026 21:28:18 GMT"
```
lukas-reining: That's a funny one :D

service/openapi.yaml
```diff
@@ -498,3 +698,24 @@ components:
     flagMetadataDescription:
       description: |
         Arbitrary metadata for the flag, useful for telemetry and documentary purposes.
+webhooks:
```
lukas-reining: I think this is wrong. Webhooks are literally meant as webhooks per the spec. 

> https://spec.openapis.org/oas/v3.2.0.html#oas-webhooks
> The incoming webhooks that MAY be received as part of this API and that the API consumer MAY choose to implement. Closely related to the callbacks feature, this section describes requests initiated other than by an API call, for example by an out of band registration. [....]

This would need to look more like this: 

```yaml
openapi: 3.2.0
info:
  title: SSE Example API
  version: 1.0.0

paths:
  /events:
    get:
      summary: Subscribe to server-sent events
      responses:
        '200':
          description: Stream of events
          content:
            text/event-stream:
              itemSchema:  # <-- This describes EACH event in the stream
                type: object
                properties:
                  id:
                    type: string
                  timestamp:
                    type: string
                    format: date-time
                  data:
                    type: string
```

I get why you put it here as we do not know the real path but I think semantically something like the following would be more suited: 

```yaml
paths:
  /{implementationPath}:
    parameters:
      - name: implementationPath
        in: path
        required: true
        schema:
          type: string
        description: Implementation-specific path (e.g., "events", "stream", "sse")
    get:
      summary: SSE Event Stream
      description: The actual path is determined by the implementation
      responses:
        '200':
          description: Stream of events
          content:
            text/event-stream:
              itemSchema:
                $ref: '#/components/schemas/Event'
```

service/openapi.yaml
```diff
@@ -308,6 +364,150 @@ components:
...
+        lastModified:
+          description: |
+            Latest flag configuration timestamp. Supports Unix timestamp in
+            seconds (recommended) or a date string (ISO 8601 or HTTP-date).
+            If present, providers should include it as the `flagConfigLastModified`
+            query parameter on the re-fetch request.
+          oneOf:
+            - type: integer
+              minimum: 0
+            - type: string
```
lukas-reining, line 507:
```suggestion
            - type: string
            - format: date-time
```
This should be fine as it is: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

lukas-reining (commented):
Looks good so far, left some feedback.

service/openapi.yaml
jonathannorris: Made `endpoint.origin` optional in the latest commit. If absent, providers should use their configured OFREP base URL origin. This way `endpoint` can be used with just `requestUri` for simple proxy setups that only need to preserve the path/query, while still supporting explicit origin override when needed.

erka (approved):

service/openapi.yaml
jonathannorris: You're right that `webhooks` is semantically wrong here.

Moved the SSE event format documentation from `webhooks` to `components/pathItems/eventStreamConnection` in the latest commit. This keeps the `text/event-stream` + `itemSchema` representation (which is the correct OAS 3.2.0 pattern per §4.14.4) without misusing `webhooks` or introducing a fake path. The description makes it clear that this is documentation-only and the actual SSE endpoints are opaque/vendor-provided via `eventStreams`.

jonathannorris:
Split the spec into two files to keep the main `openapi.yaml` at OAS 3.1.0 with full spectral linting while still documenting the SSE event format using OAS 3.2.0's `text/event-stream` + `itemSchema` support:

- `service/openapi.yaml` (3.1.0) — core OFREP spec with all schemas including `sseEvent` / `sseEventData`
- `service/event-streams.yaml` (3.2.0) — supplementary file documenting the SSE event stream format via `components/pathItems`, referencing the schemas in the main spec

lukas-reining (approved):
Looks good to me, thanks @jonathannorris!

PR #69, opened 2026-03-19
docs: add ADR 0010 to disable default polling for static-context providers

### LINKED ISSUE #68: Disable Default Polling for OFREP Static-Context Providers
## Problem

OFREP static-context providers currently poll on a fixed timer by default (JS: 30s, Swift: 30s, Kotlin: 5min). This has two issues:

1. **Sticky bucketing** — polling causes flag values to change mid-session, which breaks the expectation that a user sees consistent behavior throughout a session. Most platforms expect bucketing to remain sticky for a user's whole session.

2. **Resource efficiency** — timer-based polling is wasteful for static-context providers where the evaluation context rarely changes during a session. This is especially problematic on mobile where it impacts battery life and data usage.

This is out of step with how vendor SDKs handle client-side refresh. DevCycle, LaunchDarkly, Statsig, and Eppo do not use timer-based polling as the default client-side refresh mechanism:

| Vendor | Default Mechanism | Timer Polling? |
|--------|-------------------|----------------|
| **DevCycle** | Init + context change + foreground + SSE push | No |
| **LaunchDarkly** | SSE streaming foreground, 60min poll background only | No (streams) |
| **Statsig** | Init + user change only | No |
| **Eppo** | Init only, polling opt-in | No (off by default) |
| **ConfigCat** | Auto polling (60s) | Yes |
| **Unleash** | Polling (30s) | Yes |

## Proposed Solution

Replace default timer-based polling with event-driven refresh for static-context providers. Providers should re-fetch bulk evaluations on:

1. **Initialization** (already defined in ADR-0009)
2. **Context change** (already handled via `onContextChange`)
3. **App foreground / page visibility** — re-fetch when the app returns from background or the page becomes visible

Timer-based polling should remain available as an opt-in for applications that want it, but default to disabled.

Further details on foreground detection, debouncing, and interaction with SSE (see #62) will follow in an ADR PR.

jonathannorris:
## Summary

- Proposes disabling default timer-based polling for static-context providers
- Providers should re-fetch on initialization, context change, app foreground / page visibility, and SSE push
- Timer-based polling remains available as opt-in but defaults to disabled

## Motivation

Current OFREP static-context providers poll on a fixed timer by default (JS: 30s, Swift: 30s, Kotlin: 5min). This breaks sticky bucketing expectations and wastes resources on mobile. The majority of vendor SDKs (DevCycle, LaunchDarkly, Statsig, Eppo) do not use timer-based polling as the default client-side refresh mechanism.

See the [vendor comparison table](https://github.com/open-feature/protocol/issues/68) in the issue for details.

## Notes

- This supersedes the default polling behavior in ADR-0005 for static-context providers only. ADR-0005 remains valid for the polling mechanism itself and for server-side providers.
- Designed to work alongside ADR-0008 (SSE) and ADR-0009 (local persistence / cache-first initialization).
- Open questions around foreground refresh cooldown duration and migration path for existing applications.

### Related Issues

Closes #68

### DIFF
--- a/service/adrs/0010-disable-default-polling-for-static-context-providers.md
+++ b/service/adrs/0010-disable-default-polling-for-static-context-providers.md
@@ -12,7 +12,7 @@ OFREP static-context providers currently poll on a fixed timer by default. The J
 
 Timer-based polling is problematic for static-context providers for two reasons:
 
-1. **Sticky bucketing** — polling causes flag values to change mid-session, breaking the expectation that a user sees consistent behavior throughout a session. Most platforms expect feature flag bucketing to remain sticky for a user's whole session. Changing flag values mid-session can cause UI flicker, inconsistent experiences, and breaks experiment integrity.
+1. **Sticky bucketing** — while bucketing is generally deterministic for the same context, time-based targeting rules (commonly used in rollouts) can get triggered between polling intervals, causing evaluated values to change mid-session. Most platforms expect feature flag bucketing to remain sticky for a user's whole session. Changing flag values mid-session can cause UI flicker, inconsistent experiences, and breaks experiment integrity.
 
 2. **Resource efficiency** — timer-based polling is wasteful for static-context providers where the evaluation context rarely changes during a session. On mobile, frequent polling impacts battery life and data usage unnecessarily.
 
--- a/service/adrs/0010-disable-default-polling-for-static-context-providers.md
+++ b/service/adrs/0010-disable-default-polling-for-static-context-providers.md
@@ -86,7 +86,6 @@ All re-fetch triggers (foreground, SSE push, context change, opt-in polling) sho
 ## Implementation Notes
 
 - Foreground detection should be implemented behind a platform abstraction so it can be tested and swapped
-- Providers should apply a short cooldown (e.g., 5-10 seconds) on foreground re-fetches to avoid excessive requests from rapid foreground/background cycling
 - The existing `pollInterval` configuration should default to `0` (disabled) instead of the current positive values (JS: 30000ms, Swift: 30s, Kotlin: 5min). Applications can opt into polling by setting a positive value.
 - This ADR supersedes the default polling behavior established in [ADR-0005](0005-polling-for-bulk-evaluation-changes.md) for static-context providers only. ADR-0005 remains valid for the polling protocol mechanism itself and for dynamic-context (server-side) providers.
 - Providers should emit `PROVIDER_CONFIGURATION_CHANGED` when a foreground re-fetch returns updated flag values
@@ -94,7 +93,8 @@ All re-fetch triggers (foreground, SSE push, context change, opt-in polling) sho
 
 ## Open Questions
 
-1. Should the ADR specify a recommended cooldown duration for foreground re-fetches, or leave this to implementations?
+1. ~~Should the ADR specify a recommended cooldown duration for foreground re-fetches, or leave this to implementations?~~
+   - **Answer:** Not needed. ETag/304 already short-circuits redundant requests, and rapid foreground/background cycling is not a realistic concern.
 2. How should providers handle the transition for existing applications that depend on the current default polling behavior? Should there be a deprecation period or migration guide?
    - **Answer:** This can be treated as a breaking provider change. OFREP providers are all sub-v1, so there is no semver obligation to provide a deprecation period. Breaking provider changes are typically just changes in options or behaviors and are straightforward to handle.
 
--- a/service/adrs/0010-disable-default-polling-for-static-context-providers.md
+++ b/service/adrs/0010-disable-default-polling-for-static-context-providers.md
@@ -96,6 +96,7 @@ All re-fetch triggers (foreground, SSE push, context change, opt-in polling) sho
 
 1. Should the ADR specify a recommended cooldown duration for foreground re-fetches, or leave this to implementations?
 2. How should providers handle the transition for existing applications that depend on the current default polling behavior? Should there be a deprecation period or migration guide?
+   - **Answer:** This can be treated as a breaking provider change. OFREP providers are all sub-v1, so there is no semver obligation to provide a deprecation period. Breaking provider changes are typically just changes in options or behaviors and are straightforward to handle.
 
 ## Related
 
--- a/service/adrs/0010-disable-default-polling-for-static-context-providers.md
+++ b/service/adrs/0010-disable-default-polling-for-static-context-providers.md
@@ -51,7 +51,7 @@ Providers should detect app foreground / page visibility using platform-appropri
 
 On foreground, if the provider has an active SSE connection, no re-fetch is needed as the SSE connection handles change detection. If the SSE connection was closed during inactivity (as described in [ADR-0008](0008-sse-for-bulk-evaluation-changes.md)), the provider should reconnect and perform a full unconditional re-fetch.
 
-If SSE is not configured or not supported by the server, the provider should re-fetch on foreground unconditionally.
+If SSE is not configured or not supported by the server, the provider should re-fetch on foreground.
 
 ### Interaction with SSE (ADR-0008)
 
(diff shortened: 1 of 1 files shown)


service/adrs/0010-disable-default-polling-for-static-context-providers.md
```diff
@@ -0,0 +1,123 @@
...
+- The existing `pollInterval` configuration should default to `0` (disabled) instead of the current positive values (JS: 30000ms, Swift: 30s, Kotlin: 5min). Applications can opt into polling by setting a positive value.
+- This ADR supersedes the default polling behavior established in [ADR-0005](0005-polling-for-bulk-evaluation-changes.md) for static-context providers only. ADR-0005 remains valid for the polling protocol mechanism itself and for dynamic-context (server-side) providers.
+- Providers should emit `PROVIDER_CONFIGURATION_CHANGED` when a foreground re-fetch returns updated flag values
+- SDK documentation should note the change in default behavior and how to opt into timer-based polling if needed
+
+## Open Questions
+
+1. Should the ADR specify a recommended cooldown duration for foreground re-fetches, or leave this to implementations?
+2. How should providers handle the transition for existing applications that depend on the current default polling behavior? Should there be a deprecation period or migration guide?
```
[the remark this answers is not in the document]
toddbaert: @jonathannorris I agree with this - we can prescript this as a breaking provider change. Breaking provider changes are easy to deal with generally anyway, since they are usually just changes in options or behaviors.

Our OFREP providers are also sub v1.

toddbaert (approved):
I'm good with this, but interested in what @beeme1mr @lukas-reining @thomaspoignant think.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
```diff
@@ -0,0 +1,105 @@
...
+
+Proposed
+
+## Context
+
+OFREP static-context providers currently poll on a fixed timer by default. The JS web provider polls every 30 seconds, the Swift provider every 30 seconds, and the Kotlin provider every 5 minutes. This was originally specified in [ADR-0005](0005-polling-for-bulk-evaluation-changes.md), which established polling as the change detection mechanism for OFREP.
+
+Timer-based polling is problematic for static-context providers for two reasons:
+
+1. **Sticky bucketing** — polling causes flag values to change mid-session, breaking the expectation that a user sees consistent behavior throughout a session. Most platforms expect feature flag bucketing to remain sticky for a user's whole session. Changing flag values mid-session can cause UI flicker, inconsistent experiences, and breaks experiment integrity.
```
lukas-reining: Why does polling change the bucketing? 
Wouldn't we assume the bucketing to be consistent throughout all reloads of an app during the experiment? The polling is noting else than a reload to me.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
```diff
@@ -0,0 +1,105 @@
...
+| **ConfigCat** | Auto polling (60s) | Yes |
+| **Unleash** | Polling (30s) | Yes |
+
+DevCycle's mobile and web SDKs re-fetch configuration only on initialization, user change (`identifyUser`/`resetUser`), app foreground / page visibility, and SSE push events. There is no timer-based polling for config. LaunchDarkly uses SSE streaming when foregrounded and only polls in the background at 60-minute intervals. Statsig and Eppo do not poll at all by default.
+
+[ADR-0008](0008-sse-for-bulk-evaluation-changes.md) introduces SSE as an optional real-time change notification mechanism. Combined with event-driven refresh triggers, this provides a complete alternative to timer-based polling for static-context providers.
+
+## Decision
+
+Static-context providers should not poll on a fixed timer by default. Instead, providers should re-fetch bulk evaluations based on event-driven triggers:
```
lukas-reining, line 36: I support this!

service/adrs/0010-disable-default-polling-for-static-context-providers.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. **Initialization** — as defined in [ADR-0009](0009-localStorageForStaticContextProviders.md), providers fetch bulk evaluations during initialization (or load from cache and refresh in the background).
+
+2. **Context change** — when `onContextChange` is invoked (e.g., user switch, logout), the provider re-fetches bulk evaluations for the new context. This is already handled by existing provider implementations.
+
+3. **App foreground / page visibility** — when the application returns from the background or the page becomes visible, the provider should re-fetch bulk evaluations. This ensures values are fresh when the user returns without requiring continuous polling.
```
lukas-reining, line 42: Would't this bring back the flickering described above? 
First the old values would be there until the request has finished.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
```diff
@@ -0,0 +1,105 @@
...
+- Reduced network traffic and battery usage on mobile devices
+- Aligns with the established pattern used by the majority of vendor SDKs (DevCycle, LaunchDarkly, Statsig, Eppo)
+- App foreground refresh ensures values are fresh when the user returns from background
+- Combined with SSE (ADR-0008), provides real-time updates without the overhead of continuous polling
+
+### Negative
+
+- Flag changes made server-side won't reach the client until the next foreground event, context change, or SSE push (if configured). Applications that need mid-session updates without SSE must opt into polling.
+- Providers must implement platform-specific foreground detection, adding complexity across iOS, Android, and web
+- Breaking change for existing applications that rely on the current default polling behavior to receive mid-session flag updates
```
lukas-reining, line 84: I think these are all acceptable considering the positive arguments.

lukas-reining (approved):
Approving with some comments.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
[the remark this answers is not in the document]
jonathannorris: yea agree, since the OFREP providers are all sub-v1 this can just be a breaking provider change. Will remove the migration open question.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
jonathannorris: While bucketing is generally deterministic, there are cases where time-based rules (commonly used in rollouts) would get triggered by polling where they would generally stay consistent between sessions if polling wasn't used. With 30s polling, those changes get picked up mid-session while the user is actively using the app, which can cause the experience to shift unexpectedly. With foreground-only refresh, changes are picked up at a natural transition point when the user returns to the app. SSE-based pushes as discussed in ADR-0008 would also trigger changes, but those are explicit server-initiated notifications rather than the client blindly re-fetching on a timer.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
jonathannorris: yea there's still a potential for the values to change on foreground, but in most real-world cases it's very unlikely there is any change when the app is foregrounded since feature flag changes are fairly infrequent. The difference is also that it happens at a natural transition point when the user returns to the app rather than unpredictably while they're actively using it. Users already expect the app to refresh when they come back to it. With polling, the values can change mid-interaction which is more disruptive.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
beeme1mr: It may be worth clarify this in the doc. Initially, I had the same question as Lukas.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
beeme1mr: I'm pretty sure the React SDK allows us to opt out of this live update behavior but it would be worth confirming.

juanparadox (approved):

erka (approved):

beeme1mr (approved):

service/adrs/0010-disable-default-polling-for-static-context-providers.md
jonathannorris: yea makes sense, I'll update the ADR to clarify that it's specifically about time-based targeting rules (commonly used in rollouts) that can get triggered between polling intervals, causing the evaluated values to change mid-session.

service/adrs/0010-disable-default-polling-for-static-context-providers.md
jonathannorris: confirmed, the React SDK has `updateOnConfigurationChanged` (default `true`) which can be set to `false` at the hook or provider level to prevent re-renders on `ConfigurationChanged`. It also does smart flag-key filtering so it only re-renders if the specific flag's value actually changed.

PR #75, opened 2026-04-13
docs: add ADR 0009 for static-context cache-first local persistence

### LINKED ISSUE #65: OFREP Static-Context Provider Local Persistence

jonathannorris:
## Summary
- add ADR 0009 proposing local persistence with local-cache-first initialization for static-context OFREP providers
- on startup, providers load cached evaluations immediately so initial flag evaluations never return defaults, then refresh from the network in parallel
- define local-cache-first and cache-miss initialization paths mapped to the OpenFeature provider lifecycle (`PROVIDER_READY`, `PROVIDER_CONFIGURATION_CHANGED`)
- introduce a `cacheMode` option with `local-cache-first` (default), `network-first`, and `disabled` values so applications can opt into blocking-on-fresh-evaluation when appropriate (e.g., SPAs)
- cache key uses `hash(targetingKey)`, or `hash(cacheKeyPrefix + ":" + targetingKey)` when a `cacheKeyPrefix` is configured for multi-provider deployments

## Motivation
Every major vendor SDK (LaunchDarkly, Statsig, DevCycle, Eppo) uses local-cache-first initialization by default. Current OFREP static-context providers keep their cache in memory only, losing all state on restart. See [vendor mobile SDK caching research](https://gist.github.com/jonathannorris/4f2f63142b70719e3c6bfe8b226a0585) for a detailed comparison.

Some applications (most notably SPAs that already block rendering behind a splash screen) prefer to wait for a fresh evaluation rather than risk cached values shifting under the user. The `cacheMode: "network-first"` option supports that pattern while still using cached values as a fallback when the network is unavailable.

## Notes
This PR replaces #64 which was closed due to a branch rename. All review feedback from #64 has been addressed.

## Related
- Closes #65
- Replaces #64

## Test plan
- not applicable; documentation-only change


service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Open Questions
+
+1. Should providers support caching evaluations for multiple targeting keys (like LaunchDarkly's `maxCachedContexts`), or only retain the most recent? Multi-context caching enables instant user switching on shared devices but increases storage usage.
+2. Should the storage key include a namespace to prevent collisions when multiple OFREP providers share the same local storage origin?
+   - **Answer:** Yes. Providers should support an optional `cacheKeyPrefix` configuration option. When provided, the cache key becomes `hash(cacheKeyPrefix + targetingKey)` instead of `hash(targetingKey)`. The prefix value is left to the application author (e.g., the OFREP base URL, a project or auth token, or any other distinguishing string). The default (no prefix) keeps the single-provider case simple. See the `cacheKeyPrefix` section in the Decision above.
```
jonathannorris: cc @lukas-reining

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,217 @@
...
+### Initialization
+
+During initialization, a provider should follow a cache-first approach:
+
+1. Attempt to load a matching persisted bulk evaluation from local storage (matching `cacheKeyHash`).
+2. **If a matching persisted entry exists (cache hit):**
+   - Populate the in-memory cache from the persisted entry immediately.
+   - Return from `initialize()` so the SDK can emit `PROVIDER_READY`. Evaluations served from the persisted entry should use `CACHED` as the evaluation reason.
+   - Attempt the `/ofrep/v1/evaluate/flags` request in the background.
+   - If the background request succeeds, update the in-memory cache from the response, update the persisted entry, and emit `PROVIDER_CONFIGURATION_CHANGED`. Evaluations should switch to the server-provided reasons.
```
erka, line 77: `update the persisted entry` could fail for various reasons. It would be great to clarify the expected behavior in such cases so all implementations can follow a consistent pattern.

service/adrs/0009-local-storage-for-static-context-providers.md
erka:
```js
hash('foo' + 'bar') == hash('fo' + 'obar')
```
This might lead to collisions, so introducing an additional character in between may help avoid them.

jonathannorris: Good catch @erka. Updated to use a delimiter: `hash(cacheKeyPrefix + ":" + targetingKey)` to avoid ambiguous concatenation.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: @erka I don't think there's much we can do beyond logging the error and continuing with the fresh values in memory. The old persisted entry would remain on disk for the next cold start, which is still better than having nothing. A delete attempt could also fail for the same reasons. Will add a note to the implementation notes.

toddbaert (approved):
I approved https://github.com/open-feature/protocol/pull/64 previously. Reviewed the diff between the two; the key changes I see:

- `cacheKeyPrefix` is now folded into the hash itself 👍 
- Auth/config errors (401/403/400) no longer clear the persisted cache; TTL handles eventual expiry instead
- TTL is promoted from an open question to a recommendation
- storage write failure stuff (log and continue)

All of these seem like improvements to me. LGTM.

The [flickering](https://github.com/open-feature/protocol/pull/64#discussion_r2974239683) concern is still a notable tradeoff. cc @beeme1mr

erka (approved):

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+   - If the background request fails with a transient or server error (network unavailable, `5xx`), continue serving cached values and retry on the normal polling schedule.
+   - If the background request fails with an authorization or configuration error (`401`, `403`, `400`), surface the error via logging or provider error events and continue serving cached values for the current session. The persisted entry should not be cleared; the cache TTL is responsible for eventual expiry. This ensures that subsequent cold starts can still bootstrap from cached values while the error is investigated, rather than immediately degrading to defaults.
```
lukas-reining: In the case we can not update the cache for longer than the TTL, we still clear the cache right? 
Are we aligned that we never want to serve data that has reached its TTL?

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+3. **If no matching persisted entry exists (cache miss):**
+   - Attempt the `/ofrep/v1/evaluate/flags` request and await the response.
+   - If the request succeeds, populate the in-memory cache from the response, persist the entry, and return from `initialize()` (SDK emits `PROVIDER_READY`).
+   - If the request fails with a transient or server error, preserve the existing initialization failure behavior (SDK emits `PROVIDER_ERROR`).
+   - If the request fails with an authorization or configuration error, preserve the existing initialization failure behavior (SDK emits `PROVIDER_ERROR` with error code `PROVIDER_FATAL`).
+
+```mermaid
```
lukas-reining, line 141: Just a style question but having these nested `alt`s is very hard to read. 
Maybe we could split it into 2 or 3 diagrams. 
Maybe the case with and without cache hit could make it. 
But it is totally optional.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+The accepted tradeoff is that the cache is keyed by stable user identity: a change in `targetingKey` (user switch, logout) invalidates the cache, but changes to other context properties do not.
+Those properties only affect evaluation when the server is reachable, at which point the provider refreshes anyway.
+
+When the provider has not initialized from cache (cache miss path), providers must not silently fall back to persisted data for authorization failures, invalid requests, or other responses that indicate a configuration or protocol problem.
+
+When the provider has already initialized from cache (cache hit path), authorization or configuration errors from the background refresh should be surfaced via logging or provider error events. The provider should continue serving cached values for the current session rather than revoking a working state. The persisted entry should not be cleared on auth or config errors; the cache TTL is responsible for eventual expiry. This avoids degrading subsequent cold starts to defaults while the error is investigated.
```
lukas-reining: > The persisted entry should not be cleared on auth or config errors; the cache TTL is responsible for eventual expiry. This avoids degrading subsequent cold starts to defaults while the error is investigated.

Do we want to specify when the cache clearing happens? 
Thinking about it, it is merely an implementation detail, the only important part is that the cache entry is only used until its TTL passes.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+- SDK documentation should note that initial evaluations may return cached values (with `CACHED` reason) that are subsequently updated when fresh values arrive
+- Providers should enforce a configurable TTL on persisted entries to ensure stale caches are eventually purged, particularly in cases where the provider can no longer refresh from the server (e.g. persistent auth errors). Since auth and config errors do not clear the persisted cache, the TTL is the mechanism that prevents indefinitely stale data. DevCycle uses a 30-day default (`configCacheTTL`) as a reference.
+- If a storage write fails (e.g. quota exceeded, permission denied), the provider should log the error and continue operating with the fresh values in memory. The previously persisted entry, if any, remains on disk for the next cold start.
+
+## Open Questions
+
+1. Should providers support caching evaluations for multiple targeting keys (like LaunchDarkly's `maxCachedContexts`), or only retain the most recent? Multi-context caching enables instant user switching on shared devices but increases storage usage.
```
lukas-reining, line 297: I think they should but this is fully allowed by this proposal already, isn't it?

lukas-reining (approved):
Looks good to me! Left some minor questions.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+Date: 2026-03-06
+
+## Status
+
+Proposed
+
+## Context
+
+OFREP static-context providers evaluate all flags in one request and then serve evaluations from a local cache.
+Current implementations in `js-sdk-contrib`, `kotlin-sdk-contrib`, and `ofrep-swift-client-provider` keep that cache in memory only.
```
beeme1mr, line 12: Is the swift provider in its own repo?

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+Persistent local storage acts as the source used to bootstrap that in-memory cache on startup and update it on each successful refresh.
+
+### Initialization
+
+During initialization, a provider should follow a cache-first approach:
+
+1. Attempt to load a matching persisted bulk evaluation from local storage (matching `cacheKeyHash`).
+2. **If a matching persisted entry exists (cache hit):**
+   - Populate the in-memory cache from the persisted entry immediately.
+   - Return from `initialize()` so the SDK can emit `PROVIDER_READY`. Evaluations served from the persisted entry should use `CACHED` as the evaluation reason.
```
beeme1mr, line 75: Will the recent changes to the provider state impact this `initialize()` process? FYI @toddbaert

service/adrs/0009-local-storage-for-static-context-providers.md
beeme1mr: I feel that logging **AND** emitting a provider error makes sense.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -0,0 +1,218 @@
...
+- Web and mobile applications preserve feature state across restarts instead of losing it with the in-memory cache
+- The decision aligns with the established pattern used by vendor SDKs (LaunchDarkly, Statsig, DevCycle, Eppo) and with the existing OFREP model where static-context providers evaluate remotely once and then read locally
+- Reusing the stored `ETag` allows efficient revalidation when connectivity returns
+- Provider implementations get a consistent default expectation for offline behavior across ecosystems
+
+### Negative
+
+- Providers become more complex because they must manage persistence, cache-key matching, and recovery flows
+- Persisted evaluations may become stale, so applications can continue using outdated flag values while offline
+- Applications may briefly see stale cached values before fresh values arrive, and should handle `PROVIDER_CONFIGURATION_CHANGED` events if they need to react to updates
```
beeme1mr, line 248: This is a fairly major behavior change when you're using the OFREP provider with the web SDK. That doesn't mean we shouldn't do it, but it's something we'll need to properly document in the provider.

beeme1mr (commented):
Overall, I like the proposal. I just kept thinking about how I'd prefer it if the Provider evaluated the flags before falling back to the cached values. It would be nice if that were at least a configuration option. The reason I think this may be important is that some apps may not respond to flag changes and therefore continue using the old values for the duration of the session. Even ones that do react run the risk of the screen jitting when flag values change shortly after page load.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: yes, no mono-repos in swift.

jonathannorris:
Pushed some updates after a discussion with @beeme1mr on the network-first use case.

**Summary of changes**

For mobile and most web apps, local-cache-first is clearly the right default (eliminates the flash-of-defaults problem, matches vendor SDK behavior). But for some web use cases, particularly SPAs that already block rendering behind a splash screen, it makes more sense to block initialization on a fresh evaluation rather than flash cached values that might shift under the user.

To support both patterns, the ADR now defines a single `cacheMode` option with three values:

- **`local-cache-first`** (default): the original described behavior of the ADR, load from cache immediately and refresh in the background
- **`network-first`**: Simiar to the existing behaviour with a local-cache backup. `initialize()` awaits the network response using the existing `timeoutMs`. On success, fresh values only. On network failure, `5xx`, or timeout, fall back to the persisted cache if one exists. Auth/config errors (`401`/`403`/`400`) emit `PROVIDER_ERROR` fatal without falling back to cache
- **`disabled`**: no persistence at all, in-memory cache only, `initialize()` blocks on the network

The earlier `disableLocalCache` boolean is replaced by `cacheMode: \"disabled\"`. Since nothing has implemented the ADR yet (still Proposed), the rename is non-breaking.

**Other tweaks**

- Added \`Alternatives Considered\` entries documenting why we chose a single enum over two booleans, and why we didn't go with platform-specific defaults
- Added guidance that apps using \`network-first\` should consider lowering \`timeoutMs\` from the default so users aren't stuck on a loading state when the network is unavailable
- Clarified fallback semantics in the cache matching section: \`network-first\` only falls back to cache on network errors, never on auth errors

cc @lukas-reining @guidobrei @toddbaert @beeme1mr @erka

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: yes, TTL-expired entries should not be served — the provider should treat them as cache-miss. I'll tighten the implementation note to make this explicit.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: split into three separate diagrams: cache hit (with background refresh), cache miss, and the normal polling cycle.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: agreed, it's an implementation detail. The important invariant is that TTL-expired entries aren't served (just made that explicit in the TTL implementation note). When the actual disk cleanup happens is up to the implementation.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: yes, `cacheKeyPrefix` covers this — multi-provider deployments are supported today by configuring a distinct prefix per instance.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: agreed, updated to require both: log the error and emit a `PROVIDER_ERROR` event on auth/config errors during background refresh.

beeme1mr (approved):

lukas-reining (approved):
Looks good to me!

PR #80, opened 2026-06-19
docs: amend ADR 0009 to tie cache key to OFREP resource (domain, url, auth)

jonathannorris:
Amends [ADR 0009](https://github.com/open-feature/protocol/blob/main/service/adrs/0009-local-storage-for-static-context-providers.md) (static-context local persistence) to tie the persisted cache key to the OFREP resource the evaluation was fetched from, by including the provider's bound `domain`, the OFREP base URL, and the auth credential, in addition to the `targetingKey`.

## Motivation

The accepted ADR keys the persisted cache on `targetingKey` (optionally prefixed with an application-supplied `cacheKeyPrefix`), leaving collision avoidance up to the application. A persisted evaluation is really a function of the resource it came from (server + credential + bound domain) and the identity it was requested for, so the cache key should reflect that automatically.

Surfaced in [open-feature/js-sdk-contrib#1566](https://github.com/open-feature/js-sdk-contrib/pull/1566), where the OFREP web provider's local persistence has no way to know which `domain` it's bound to.

## Changes

- Cache key becomes `hash(url + ":" + auth + ":" + domain + ":" + targetingKey)` (with optional `cacheKeyPrefix` still prepended). `cacheKeyPrefix` is demoted to a supplement for app-controlled namespacing.
- Updated the Decision, cache-matching, example payload, and Open Questions #2/#3/#4 to match.
- The bound `domain` relies on [open-feature/spec#393](https://github.com/open-feature/spec/pull/393), which supplies the `domain` to `initialize` and adds an opt-in `domain-scoped` declaration the API enforces, so a persisting OFREP provider is bound to a single `domain` and the `domain` component of the key is unambiguous.

## Open for discussion

Two points flagged as open questions in the ADR:

- **Including the auth credential** (Open Question #3) is the piece most open to discussion. OFREP supports rotating/short-lived tokens via `headersFactory`, and a token that rotates on every request would defeat persistence. But auth doesn't change on every request, so for the common case of a stable credential it still provides useful separation between projects/environments/keys against the same URL. The original ADR dropped `authToken` for the rotation reason (see [protocol#64](https://github.com/open-feature/protocol/pull/64)); reviewers may prefer a stable credential identifier or making the auth component opt-in.
- **Removing `cacheKeyPrefix`** (Open Question #4): now that the URL, auth, and `domain` are in the key by default, the collisions `cacheKeyPrefix` was added to solve are handled automatically. I lean toward removing it to simplify the config surface, but open to keeping it as an explicit escape hatch.

Marked as a draft for discussion.


toddbaert (approved):

JamieSinn (approved):

MattIPv4 (commented):
LGTM, but is there interest in discussing the transformer approach I proposed in https://github.com/open-feature/js-sdk-contrib/pull/1566 -- that seems even more versatile than this proposed update, while still solving for the concerns that originally landed on just having targeting key to avoid volatility?

jonathannorris:
> open-feature/js-sdk-contrib#1566

@MattIPv4 we can look at that separately to this, would likely need an ADR change as well. I think this change to include the `url` + `auth` + `domain` is more important to try and fix the default behaviour. I guess the main question would be what's the use-case that needs a transformer when the hash is: `hash(url + ":" + auth + ":" + domain + ":" + targetingKey)` with or without the `cacheKeyPrefix`.

MattIPv4:
👍 That's fair, can be looked into after. I can't think of an immediate use case (domain solves our issue), but it seems like a nice bit of flexibility to ensure folks have full control of the caching rather than the caching making assumptions (which is what caused this change to be needed).

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -297,3 +307,10 @@ A single default (local-cache-first) with an explicit per-application opt-out is
...
+   - **Auth header:** Including the auth credential would tie the cache even more tightly to the protected resource, but credentials are not always stable. OFREP supports rotating/short-lived tokens via `headersFactory`, and a rotating bearer token would change the hash on every rotation. This is the same reason the original ADR dropped `authToken` from the cache key (see the [protocol#64](https://github.com/open-feature/protocol/pull/64) discussion).
+   - **Proposed (2026-06-19):** Include all three (OFREP base URL, auth credential, and bound `domain`) in the cache key; the Decision section above reflects this. The auth credential is the most open to discussion: it is less effective for short-lived or rotating tokens, but auth does not change on every request, so it still provides useful separation for the common case of a stable credential.
+4. **(Amendment, 2026-06-19)** Should the optional `cacheKeyPrefix` configuration option be removed entirely?
```
beeme1mr: I don't think it's worth adding amendment dates throughout the doc. Something at the top or bottom of the file is sufficient IMO.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -279,21 +288,29 @@ A single default (local-cache-first) with an explicit per-application opt-out is
...
+   - **OFREP URL** Folding the OFREP base URL into the cache key (e.g. `hash(url + ":" + domain + ":" + targetingKey)`) ties cached results directly to the resource that produced them, so a provider reconfigured to point at a different server does not serve another server's cached evaluations, and same-origin instances pointing at different servers separate automatically without an explicitly configured prefix. The base URL is stable across restarts, so it does not reintroduce the volatile-input problem. This mirrors vendor SDKs that key their persisted cache by SDK key or environment (Statsig, Eppo, ConfigCat). The cost is that changing the configured URL silently invalidates the cache, which is usually the desired behavior.
+   - **Auth header:** Including the auth credential would tie the cache even more tightly to the protected resource, but credentials are not always stable. OFREP supports rotating/short-lived tokens via `headersFactory`, and a rotating bearer token would change the hash on every rotation. This is the same reason the original ADR dropped `authToken` from the cache key (see the [protocol#64](https://github.com/open-feature/protocol/pull/64) discussion).
```
askpt, line 331: This open question is very provider-dependent. I would prefer to get it removed from the key. If a provider has a refresh of 10 minutes, that would mean the cache is invalidated quite often. 

Wouldn't make more sense to tie to the `ETag` instead of auth header?

askpt (commented):
Minor comment. The rest looks good.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -196,12 +205,12 @@ If the background refresh fails and the provider cannot confirm that cached valu
...
+The cache key is intentionally derived from the OFREP resource and `targetingKey` rather than the full evaluation context.
 Static-context evaluations on the server can depend on context properties beyond `targetingKey`, so cached values may not reflect the current full context.
 However, hashing the full context is impractical for local-cache-first startup because many implementations set volatile context properties on initialization (e.g. `lastSessionTime`, `lastSeen`, `sessionId`) that would change the hash on every app restart, defeating the purpose of persistence.
-The accepted tradeoff is that the cache is keyed by stable user identity: a change in `targetingKey` (user switch, logout) invalidates the cache, but changes to other context properties do not.
+The accepted tradeoff is that the cache is keyed by stable inputs (the OFREP base URL, the auth credential, the bound `domain`, and the `targetingKey`): a change in `targetingKey` (user switch, logout), in the bound `domain`, or in the resource the provider points at invalidates the cache, but changes to other context properties do not.
```
lukas-reining: My largest concern is that the auth credential might not be "stable" in that sense e.g. with a JWT of 5 minutes expiry. 
This might be okay though. 
What I think could become complicated is when an app starts it might try to fetch a new token first, then fail do the evaluation and be stuck with a new credential but wanting to use the cache.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -279,21 +288,29 @@ A single default (local-cache-first) with an explicit per-application opt-out is
 
 - "Local storage" means a local persistent key-value store appropriate for the runtime, such as browser `localStorage` on the web or an equivalent mobile storage mechanism
 - Providers should version their persisted format so future schema changes can be handled safely
+- Persisting providers should declare themselves `domain-scoped` (per [open-feature/spec#393](https://github.com/open-feature/spec/pull/393)) so the API binds each instance to at most one `domain`. This keeps the `domain` component of the cache key unambiguous and avoids a single shared instance writing entries for more than one `domain`
```
lukas-reining, line 309: Does this include the default domain? So it means it should implement the optional parameter?

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -279,21 +288,29 @@ A single default (local-cache-first) with an explicit per-application opt-out is
...
-- Providers should enforce a configurable TTL on persisted entries to ensure stale caches are eventually purged, particularly in cases where the provider can no longer refresh from the server (e.g. persistent auth errors). Since auth and config errors do not clear the persisted cache, the TTL is the mechanism that prevents indefinitely stale data. A persisted entry past its TTL must not be served to the application: the provider should treat it as a cache miss and fall through to the cache-miss path. DevCycle uses a 30-day default (`configCacheTTL`) as a reference.
+- Providers should enforce a configurable TTL on persisted entries to ensure stale caches are eventually purged, particularly in cases where the provider can no longer refresh from the server (e.g. persistent auth errors). Since auth and config errors do not clear the persisted cache, the TTL is the mechanism that prevents indefinitely stale data. A persisted entry past its TTL must not be served to the application: the provider should treat it as a cache miss and fall through to the cache-miss path. DevCycle uses a 30-day default (`configCacheTTL`) as a ref erence.
```
lukas-reining:
```suggestion
- Providers should enforce a configurable TTL on persisted entries to ensure stale caches are eventually purged, particularly in cases where the provider can no longer refresh from the server (e.g. persistent auth errors). Since auth and config errors do not clear the persisted cache, the TTL is the mechanism that prevents indefinitely stale data. A persisted entry past its TTL must not be served to the application: the provider should treat it as a cache miss and fall through to the cache-miss path. DevCycle uses a 30-day default (`configCacheTTL`) as a reference.
```

Is this a typo?

lukas-reining (commented):
Looks good to me. Only the auth token is still open to me.

beeme1mr (approved):

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -34,17 +36,24 @@ The persisted entry should include:
...
+The cache key is tied to the OFREP resource the evaluation was fetched from and the identity it was requested for, so it includes:
+
+- the **OFREP base URL**, so a provider pointed at a different server does not serve another server's cached evaluations.
+- the **auth credential**, so evaluations fetched under different credentials (projects, environments, or keys against the same URL) do not collide.
+- the provider's bound **`domain`**, OpenFeature's intended unit of isolation between provider instances (the binding name passed to `setProvider`), supplied to `initialize` per [open-feature/spec#393](https://github.com/open-feature/spec/pull/393). It is empty when a provider has no bound `domain`.
+- the **`targetingKey`**, keying the entry to the user identity (see "Cache matching and fallback" below).
+
+Providers should additionally support an optional `cacheKeyPrefix` option, prepended to the hash (`hash(cacheKeyPrefix + ":" + url + ":" + auth + ":" + domain + ":" + targetingKey)`) so applications can namespace across storage partitions they control directly. The prefix can be any distinguishing string.
```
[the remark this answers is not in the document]
jonathannorris: the `+ ":" +` is just illustrative shorthand; the exact encoding is an implementation detail left to each provider. In the JS provider we serialize the fields with JSON to join them, which avoids the delimiter-collision issue.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: agreed the auth credential isn't always stable, I called that out in the open question. on ETag: it can be used differently depending on the provider, but generally it's a caching/version identifier rather than a resource identifier, so I'm not sure it fits as a key component. you'd also want the local cache to survive across ETag versions, keying on it would mean every config change misses the existing entry and defeats the persistence. it also can't be computed offline on a cold start to find the right entry, and we already persist it alongside the entry as a staleness signal.

the real question is what makes the better default. including auth almost guarantees isolation: two different credentials (projects, environments, keys) against the same url + domain can't read each other's cache without the app configuring anything. dropping it makes the default looser, that isolation is no longer guaranteed unless the app sets a `cacheKeyPrefix`. the cost of including it is that a short-lived credential (a JWT rotating every few minutes) fragments and effectively invalidates the cache on every rotation.

so it comes down to how common very low refresh times are for OFREP static-context providers in practice. if rotating-per-request tokens are the exception, the safer default seems worth it. a middle ground is keying on a stable credential identifier rather than the raw token, or making the auth component opt-in. curious where you both land.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: replied on André's thread with the same theme: the tradeoff is a safer default (auth in the key guarantees cache isolation between credentials) vs the fragmentation risk you're describing with short-lived tokens. your app-startup scenario (fetch new token, evaluation fails, stuck wanting the old cache) is the sharp edge of that. let's converge there.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: the default provider has no bound `domain`, so `initialize` is called without one and the `domain` component of the key is just empty, which is a valid single scope. a `domain-scoped` provider isn't required to have a domain, but per spec#393 (req 2.4.4) it must accept the `domain` at init, so yes, to actually scope its cache it needs to implement that optional parameter and use it.

MattIPv4 (approved):

service/adrs/0009-local-storage-for-static-context-providers.md
toddbaert: I think perhaps a lambda for a key generation might still be the best method, but by default I would do "maximum scoping" and allow people to narrow it from there. The worst thing that can happen is caches trampling each-other as we saw happen with @MattIPv4 ... between that and performance loss, I would default to maximum isolation and allow users (and wrapping providers) to relax the caching to their needs.

WRT Etag:  I think it's a poor cache key ingredient. It's a cache _validator_, not a key. A cache key should be "assemblable" before any response is received - to put it another way, it should be based on PARAMs to a request, not RESULTS of a request, or key caching use-cases become impossible.

service/adrs/0009-local-storage-for-static-context-providers.md
toddbaert: My 2c: https://github.com/open-feature/protocol/pull/80#discussion_r3520901610

service/adrs/0009-local-storage-for-static-context-providers.md
MattIPv4: > I think perhaps a lambda for a key generation might still be the best method, but by default I would do "maximum scoping" and allow people to narrow it from there

What do you mean by "maximum scoping"?

I had a lambda approach in https://github.com/open-feature/js-sdk-contrib/pull/1566, where by default the entire context is used. This guarantees maximum safety out of the box, so no trampling by default, and folks have to opt in to having a less specific key that introduces risk.

toddbaert: > What do you mean by "maximum scoping"?

I mean that the out-of-the-box defaults should add as many factors the to cache key as possible, and if they are un-needed or un-wanted in some cases, people can relax them; that's "fail safe but slow" in worse case scenarios.

MattIPv4: Ah, perfect -- that's the same way I've been thinking about this.

askpt: No concerns to me then.

askpt (approved):

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: sounds good, will update this PR to reflect that.

service/adrs/0009-local-storage-for-static-context-providers.md
```diff
@@ -297,3 +325,10 @@ A single default (local-cache-first) with an explicit per-application opt-out is
...
+   - **Answer:** Include all three (OFREP base URL, auth credential, and bound `domain`) in the default cache key, following the principle of maximum isolation out of the box: the default scopes as tightly as possible, and applications relax it where they do not need it. The auth credential is included by default; setups with rotating or short-lived credentials can drop it via the cache-key generator (Open Question #4) rather than weakening the default for everyone.
+4. Should the optional `cacheKeyPrefix` configuration option be removed entirely?
+   - **Answer:** Yes, removed. `cacheKeyPrefix` is replaced by a more general cache-key generator function. The generator receives the request inputs (OFREP base URL, auth credential, bound `domain`, `targetingKey`, and the full static context) and returns the key material the provider hashes into `cacheKeyHash`. This subsumes prefixing (prepend a namespace string) and additionally lets applications narrow the key (e.g. drop the auth credential for rotating tokens) or broaden it (e.g. add stable context fields). The provider always controls hashing; the generator only returns the value to hash.
```
MattIPv4, line 334: This makes this amendment a breaking change for providers who have already implemented this. Are we comfortable with that?

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: implemented the change with the full hash as the default + cache key generator function to override.

service/adrs/0009-local-storage-for-static-context-providers.md
jonathannorris: seeing that OFREP is still pre-1.0 I believe we are generally.

PR #82, opened 2026-07-20
docs: Amend ADR-0008 to adjust the SSE wire format to allow for customized formatting

JamieSinn:
## This PR
In cases where there's different object formats (Ably) the default format will conflict and not be valid/acted on by SDKs. This resolves that and ensures that providers can implement their own wrapping/deserialization as long as it returns the correct format to the parsed data.

For example, this also allows for varying date formats for the lastModified values.

### Related Issues

https://github.com/open-feature/js-sdk-contrib/pull/1583

### DIFF
--- a/service/adrs/0008-sse-for-bulk-evaluation-changes.md
+++ b/service/adrs/0008-sse-for-bulk-evaluation-changes.md
@@ -71,7 +71,7 @@ The `eventStreams` field is an array to support vendors whose infrastructure may
 
 ### SSE Event Format
 
-Events use the standard [SSE event format](https://html.spec.whatwg.org/multipage/server-sent-events.html). The SSE `data` field is a raw string (multiple `data:` lines are concatenated per the W3C spec) that providers must parse as JSON:
+Events use the standard [SSE event format](https://html.spec.whatwg.org/multipage/server-sent-events.html). The SSE `data` field is a raw string (multiple `data:` lines are concatenated per the W3C spec). By default, providers parse the `data` field as JSON with the shape shown below. This is the default wire format, not a fixed requirement — see [Customizable payload parsing](#customizable-payload-parsing):
 
 ```
 id: evt-1234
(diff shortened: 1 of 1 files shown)


service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -83,6 +83,16 @@ The SSE envelope `event:` field is always `message`. Using a named SSE event typ
...
+
+The default parsing behavior is: JSON-parse the SSE `data` field when it is a string, otherwise pass the value through as-is (some SSE client implementations deserialize the payload before handing it to the application). Endpoints that emit the OFREP event payload directly need no customization.
+
+Overriding the parsing is optional and is aimed at shared, extensible provider implementations reused across multiple vendors, rather than being required of every provider. Such implementations may expose a hook that receives the raw SSE message event and returns the parsed OFREP event object, defaulting to the JSON-parse-string-otherwise-passthrough behavior above. This keeps the common case zero-configuration while letting integrations with vendors that wrap or re-encode the payload adapt the wire format without changing the rest of the provider's SSE handling.
+
```
toddbaert, line 95: I don't like leaving "history" in ADRs; that's what git is for. I would just amend the edit date (at the top I believe) and talk about current state / why we want this, not what once was.

toddbaert (approved):
nit: https://github.com/open-feature/protocol/pull/82/changes#r3615615699

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: Will amend.

jonathannorris (approved):
Just update the title to `docs:` instead of `feat:`

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -83,6 +83,16 @@ The SSE envelope `event:` field is always `message`. Using a named SSE event typ
 
 Providers must inspect `data.type` to determine behavior — not the SSE envelope `event:` field.
```
lukas-reining: This `must` seems wrong with this change.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
```diff
@@ -83,6 +83,16 @@ The SSE envelope `event:` field is always `message`. Using a named SSE event typ
...
 
+#### Customizable payload parsing
+
+The JSON shape shown above is the default wire format, not a fixed requirement. The protocol does not bind providers to one specific on-the-wire encoding, because a single encoding does not hold across all SSE vendors. Some hosted SSE services do not support emitting arbitrary application payloads as the top-level `data` value and instead wrap the payload inside an additional envelope layer — for example, Ably nests the OFREP payload under `data.data`. Others may deliver the payload in a non-JSON format.
+
+What the protocol requires is that the provider can obtain an OFREP event object with `type` (and optional `etag` / `lastModified`) from each SSE message and act on it. How the raw `data` is decoded into that object is an implementation detail that may be overridden, so the SDK is not bound to a single wire encoding.
```
lukas-reining: Maybe we can reference the spec here to include the fields instead of mentioning them? 
https://github.com/open-feature/protocol/blob/7d1c3dc37090d95db590a1201d9ad0c40e7108db/service/openapi.yaml#L476

lukas-reining (approved):
This is a good addition. 
Initially I was worried that this could open the gates for many custom implementations that can become hard to get together later but I think the way it is specified does not have this problem.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: So I was 50/50 on the edit here too, but given the default is still applied as is, I wasn't sure whether or not to edit this to be more permissive, but this does accurately describe the workflow for the base implementation.

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: That's a good idea - I'll update the line ref

service/adrs/0008-sse-for-bulk-evaluation-changes.md
JamieSinn: Tweaked it a bit in language to say that it must inspect the parsed object

lukas-reining (approved):
Looks good to me!

